PR #22, opened 2024-08-28
Set up test framework.

cyon1c:
This adds the Unity Test Framework to the "native" platform, which will run tests natively on your machine using your machines g++ compiler. I found that the command `platfromio test` is the best way to execute these tests. I had trouble with the built in command `pio test` not being able to locate g++.


FinallyEve (approved):
this is def a LGTM, so we are basically stubbing out the frameworks main pass/fail mechanic? I like it! testing == less debug in the desert

PR #36, opened 2024-12-15
Add basic P2P wireless capabilities and a remote player discovery

NeoSkye:
This adds a system for sending and receiving data to other nearby PDN devices using ESP-NOW protocol. While ESP-NOW only supports packets up to 250 bytes, this change also adds a lightweight higher level protocol on top of ESP-NOW to allow for much larger packets. Note that this currently uses a system where if any part of a larger packet is dropped, the entire packet is dropped, so its still best to keep packets as small as possible. A further improvement would be to implement optional packet acknowledgement.

Services can use this protocol by adding a new packet type to the esp-now-comms.hpp header and then registering a handler to receive packets of that type. RemotePlayerManager serves as an example of how to create and register a service.

This change also supports tracking RSSI for other devices, which is the wireless signal strength between them. This can be roughly used as distance between players, though normal radio interference such as walls applies, so it's not exact. This adds a lot of extra compute required, so has been disabled by default. To enable it, set PDN_ENABLE_RSSI_TRACKING to 1 in platformio.ini

This has not been tested with recent changes due to hardware issues on my side and the large packet support has never been tested.

### DIFF
--- a/include/wireless/esp-now-comms.hpp
+++ b/include/wireless/esp-now-comms.hpp
@@ -1,0 +1,113 @@
+#pragma once
+
+#include <vector>
+#include <queue>
+#include <unordered_map>
+#include <esp_now.h>
+
+//Change to 1 to enable tracking rssi for peers
+//This works, but requires wifi to be in promiscuous mode
+//which likely prevents connecting to access points and
+//requires an unknown but likely high amount of processing
+//power
+#define PDN_ENABLE_RSSI_TRACKING 0
+
+//Use this mac address in order to reach all nearby devices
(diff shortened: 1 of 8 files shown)


NeoSkye:
Can someone please test this branch to make sure it doesn't stop normal game use before it gets merged? I'll continue trying to get my PDNs running in the meantime but I don't want this to break main branch.

NeoSkye:
I got my PDNs working and then was missing the remote player discovery messages. This turned out to be a logging issue, so switched over to use ESP_LOG style logging which is just better. Confirmed that this branch works with the existing game code and is ready for merge.

PR #45, opened 2025-04-10
Finally eve/hacking and lockdown queue disable wifi wip

FinallyEve:


src/main.cpp
```diff
@@ -37,40 +37,32 @@ void setup() {
...
     ESP_LOGI("PDN", "HW and Game Initialized\n");
 
-  // Initialize WiFi for ESP-NOW use
-  WiFi.begin();
-  // STA mode is required for ESP-NOW
-  WiFi.enableSTA(true);
-  // This could be any Wi-Fi channel, the only requirement is that all devices
-  // communicating together on ESP-NOW must use the same channel
-  WiFi.channel(6);
+    // Initialize wireless manager
```
FinallyEve, line 40: @cyon1c  this is the file that im sus about which is causing the PDNs to crash

Moroz6690 (approved):

PR #54, opened 2026-01-12
Arduino Type Extraction - First Pass for Native Emulator

cyon1c:
Summary of Changes:

efeae597986a7fc96564380d01e3951dc38a69af

- General project Clean-up.
- Resolve irregular usage of std namespace.
- Remove usages of Arduino String and other arduino types.
- Refactored Light-Strip interface
-- Removed unnecessary templating
-- Actually hide FastLED behind an interface.

9348cfc18de22c9261a9356e84a69672c39c1b1c

- Forked UUID by Rob Tillaart and removed the arduino headers.

a0fcc2954bd676998acf93a15c0068240912078a

- Created a PlatformClock interface to abstract time functions.

e8b7914fd71ebf75d1b0f5eed41c8fee661699ed

- Created a PeerCommsInterface to hide the ESP-NOW implementation.

19bbd1ec7299507f077f9b0e6247373b1cc0d661

- Refactored Wireless State Management
- Switching between ESP-NOW and wifi was unnecessary and added extra complexity.
- Created an HTTPClientInterface to hide the wifi implementation.

### DIFF
--- a/include/device/drivers/button.hpp
+++ b/include/device/drivers/button.hpp
@@ -3,8 +3,9 @@
 //
 #pragma once
 
-//Button Resources
+#include <functional>
 
+//Button Resources
 enum class ButtonIdentifier {
     PRIMARY_BUTTON = 0,
     SECONDARY_BUTTON = 1,
--- a/include/game/match-manager.hpp
+++ b/include/game/match-manager.hpp
@@ -30,7 +30,9 @@ struct ActiveDuelState {
 
 class MatchManager {
 public:
-    static MatchManager* GetInstance();
+
+    MatchManager();
+    ~MatchManager();
     /**
      * Creates a new active match
      * @param match_id UUID of the match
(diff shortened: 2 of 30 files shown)


tire-fire:
I used claude to find some still incomplete abstractions: 
PlatformClock not used everywhere: src/game/quickdraw-states/duel-state.cpp and src/game/match-manager.cpp still use millis() directly.

Arduino random() not replaced everywhere: src/game/quickdraw-states/fetch-user-state.cpp and src/game/quickdraw-states/upload-matches-state.cpp still use random(). The randomInt() helper exists in animation files but isn't in a shared location these files can use.

Arduino WIFI still used: src/wireless/quickdraw-wireless-manager.cpp directly calls WiFi.getMode(), WiFi.mode(), and EspNowManager::GetInstance()->sendData() instead of using PeerCommsInterface.

PR #68, opened 2026-02-09
Fix inverted text in DRAW image and hidden caption text

FinallyEve:
## Summary
- **Opaque image rendering**: `decodeXBMToBuffer` now writes both ON and OFF pixels, so overlaid images (like the duel DRAW screen) properly clear underlying pixels. This makes dark text like "NOW!" visible on the white box instead of bleeding through the IDLE image underneath.
- **Image captions**: Added a `name` field to the `Image` struct. The native display driver adds it to text history when present, so CLI captions show "3", "2", "1", "DRAW!" during countdown instead of stale idle-state text. Caption strings added for CONNECT, COUNTDOWN_THREE/TWO/ONE, DRAW, WIN, and LOSE across all 4 allegiance collections.

## Test plan
- [x] 3 new tests added (83 total, all passing)
- [x] Run simulator with 2 devices, cable them, trigger countdown
- [x] Verify braille mirror shows "NOW!" as dark text on white box in duel state
- [x] Verify captions show "3", "2", "1", "DRAW!" during countdown
- [x] Verify `display on`, `display off`, `mirror`, `captions` commands still work

### DIFF
--- a/include/device/drivers/native/native-display-driver.hpp
+++ b/include/device/drivers/native/native-display-driver.hpp
@@ -187,6 +187,9 @@ public:
 
     Display* drawImage(Image image, int xStart, int yStart) override {
         currentImage_ = image;
+        if (image.name) {
+            addToTextHistory(image.name);
+        }
         int x = image.defaultStartX;
         int y = image.defaultStartY;
         if (xStart != -1) x = xStart;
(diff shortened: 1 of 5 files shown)


FinallyEve:
bug with draw + inverted images now fixed, should have just used inverted whitespace from the beginning. added a place to write text to that only hits captions so we dont have cases where we draw things on the screen but its not in the captions.

PR #76, opened 2026-02-10
StateMachineManager: pause/resume lifecycle for swappable games

### LINKED ISSUE #74: Feature C: StateMachineManager — Design Input Needed
## Context

We're building the **Challenge Device System** — NPC game nodes that players cable into to play mini-games (starting with Signal Echo / Simon Says). PR #73 implements the foundational pieces:

- **Feature A**: NPC device infrastructure (ChallengeGame state machine, magic codes, NVS mode switching)
- **Feature B**: Signal Echo game (6-state MiniGame with difficulty presets, standalone playable via `--game signal-echo`)
- **Feature E**: Player profile extensions (Konami progress tracking, color profile eligibility, server sync)

These three features work independently. What's missing is the **glue** — when a player cables into an NPC device mid-Quickdraw, something needs to **pause Quickdraw, load Signal Echo, and resume Quickdraw when the game finishes**.

That something is the **StateMachineManager**, and we'd like your input on how to design it.

## What we need

A class that manages swapping between StateMachines at runtime. The player's device starts running Quickdraw. When it detects an NPC (via `cdev:` serial messages), it needs to:

1. **Pause** the current StateMachine (Quickdraw)
2. **Load** a MiniGame (Signal Echo) as the active StateMachine
3. **Run** the MiniGame until completion (`game->isGameComplete()` returns true)
4. **Resume** the previous StateMachine (Quickdraw)

## Our proposal (just a starting point — you decide)

```cpp
class StateMachineManager {
public:
    void setDefaultStateMachine(StateMachine* sm);
    void pauseAndLoad(StateMachine* sm);
    void resumePrevious();
    void loop();  // calls getActive()->loop() + checks for game completion
    StateMachine* getActive();
    bool hasPendingSwap();
};
```

Usage in the main loop would replace `game->loop()` with `smManager->loop()`.

## Questions for you

1. **Where should this class live?** We were thinking `include/state/state-machine-manager.hpp` alongside the existing `state-machine.hpp`, but you may have a better home for it.

2. **How should game-complete be communicated?** Our MiniGame base class has `isGameComplete()` which returns true when the game's outcome is set (win or lose). Should the SM Manager poll this, or would you prefer a callback/observer pattern?

3. **How should "previous SM resumed" be signaled?** When `resumePrevious()` is called, does the caller need to know it happened? (callback? flag? return value?)

4. **Ownership model**: Should SM Manager own the StateMachines (and delete them), or just hold pointers? Currently Quickdraw is owned by `main.cpp` and MiniGames would be created/destroyed per challenge encounter.

5. **Anything else we're not thinking about?** You have the full architecture picture — if there's a better pattern for this, we're all ears.

## Integration points (Feature D, blocked on this)

Once the SM Manager exists, Feature D wires everything together:

- `DeviceInstance` holds a `StateMachineManager*` instead of a raw `StateMachine*`
- CLI main loop calls `smManager->loop()` instead of `game->loop()`
- Quickdraw's Idle state detects `cdev:` → triggers `smManager->pauseAndLoad(signalEcho)`
- Signal Echo win/lose signals completion → SM Manager polls and calls `resumePrevious()`
- ESP-NOW game protocol flows between player and NPC during the mini-game

## Related

- Tracking issue: #72
- Implementation PR (A, B, E): #73
- Full plan document is in #72 comments

No rush — we'll adapt Feature D to whatever interface you build.

FinallyEve:
## Summary

Adds the StateMachineManager — a generic coordinator for pausing the default state machine and loading temporary swappable games.

**Zero external dependencies.** The SM Manager defines its own contract:
- `SwappableOutcome` — lightweight result struct (int result, int score)
- `SwappableGame` — abstract interface extending StateMachine with `isReadyForResume()`, `getOutcome()`, `getGameId()`
- `StateMachineManager` — pause/load/resume lifecycle, auto-resume on completion, outcome storage

The test fixture (`TestSwappableGame`) implements `SwappableGame` directly — no game-specific types needed. The 8 tests verify the full lifecycle in isolation.

Concrete game classes (e.g. MiniGame for the challenge device system) will extend `SwappableGame` in a follow-up PR.

Fixes #74

## How to Test

```bash
pio test -e native_cli_test
# Verify: 8 new tests pass (DefaultLoop, PauseAndLoad, AutoResume,
# OutcomeStored, ResumeToCorrectState, DeletesSwappedGame,
# PauseAndLoadWon, PauseAndLoadLost)
```

## Test plan

- [x] `pio test -e native_cli_test` — 88 tests pass (80 existing + 8 new)
- [ ] CI: Linux GCC build passes

### DIFF
--- a/include/game/test-swappable-game.hpp
+++ b/include/game/test-swappable-game.hpp
@@ -1,0 +1,97 @@
+#pragma once
+
+#ifdef UNIT_TEST
+
+#include "state/state-machine-manager.hpp"
+#include "state/state.hpp"
+
+/*
+ * TestSwappableGame — a trivial 2-state game for testing the SM Manager.
+ *
+ * State 0: TestGameIntro — waits until triggerComplete() is called.
+ * State 1: TestGameComplete — sets outcome and signals readyForResume.
+ *
+ * This class is guarded by UNIT_TEST and should never appear in production builds.
+ */
+
+class TestSwappableGame;  // Forward declaration
+
+static const int TEST_GAME_INTRO_STATE_ID = 200;
+static const int TEST_GAME_COMPLETE_STATE_ID = 201;
+static const int TEST_GAME_ID = 99;
+
+class TestGameIntro : public State {
+public:
+    TestGameIntro() : State(TEST_GAME_INTRO_STATE_ID) {}
+    void onStateMounted(Device* PDN) override {}
+    void onStateLoop(Device* PDN) override {}
+    void onStateDismounted(Device* PDN) override {}
+};
+
+class TestGameComplete : public State {
+public:
+    TestGameComplete(TestSwappableGame* game, int resultToSet) :
+        State(TEST_GAME_COMPLETE_STATE_ID),
+        game_(game),
+        resultToSet_(resultToSet)
+    {
+    }
+
+    void onStateMounted(Device* PDN) override;
+    void onStateLoop(Device* PDN) override {}
+    void onStateDismounted(Device* PDN) override {}
+    bool isTerminalState() override { return true; }
+
+private:
+    TestSwappableGame* game_;
+    int resultToSet_;
+};
+
+class TestSwappableGame : public SwappableGame {
+public:
+    TestSwappableGame(Device* device, int resultToSet = SwappableOutcome::WON) :
+        SwappableGame(device),
+        resultToSet_(resultToSet)
+    {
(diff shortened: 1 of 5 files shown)


FinallyEve:
Superseded by PRs #83/#84/#85 (rewritten on top of Cyonic's AppConfig architecture from PR #81)

PR #81, opened 2026-02-11
Implement state machine management to allow for multiple game experiences

cyon1c:
Changes:

- Added typealias AppConfig, which allows configuring what games are loaded into the pdn at run time.
- Refactored StateMachine to extend state so that it can be managed in the same way states are.
- Added `onStatePaused` and `onStateResumed` to the State API.
- Implemented extensible type `Snapshot` which allows states to implement their own data bundle to store relevant state data.
- onStatePaused returns a Snapshot, and onStateResume consumes that same snapshot
- State Machine's can only ever have 1 paused state and a single bundle for that state.
- States that are paused and then resumed will call onStateMounted before resuming, and onStateDismounted after pausing to ensure resources and peripherals are torn down correctly.

One large caveat here - in the current architecture is technically one giant memory leak since each state-machine instantiates all states on initialization, and is never garbage collected. Will look at implementing an AppHistoryStack to manage collecting apps  and simounting them when they are too far back in the app history, but for now with only a few experiences, I'm not concerned about it.


FinallyEve (approved):
LGTM, I've got my team going through and auditing their own internal PR review process to see if they can spot anything, the orchestrator will throw a comment here when shes got it with an all clear, your call on if you want to wait or go ahead and merge.

FinallyEve:
## Code Review + Embedded Performance Review (v2)

Hey Cyonic! Great improvements in the second commit — the error handling guards, `std::move`, typo fix, and 14 new device tests are exactly what was needed. Architecture is solid and our FDN stack (#83-#89) is fully compatible.

**Recommendation: APPROVE with 1 required fix, 2 suggested fixes**

---

### Must Fix

**1. `LOG_E` format string — undefined behavior** (`device.cpp`)

`currentAppId` and `appId` are `StateId` structs, not `int`. Passing a struct to `%d` is UB:

```cpp
// Current (UB):
LOG_E(TAG, "App %d not found", currentAppId);
LOG_E(TAG, "App %d not found", appId);

// Fix:
LOG_E(TAG, "App %d not found", currentAppId.id);
LOG_E(TAG, "App %d not found", appId.id);
```

This will likely cause GCC warnings on CI (Linux build).

---

### Should Fix

**2. `onStateChange()` removal drops serial flush**

The old `PDN::onStateChange()` called `lightManager->clear()` and `flushSerial()` on every state transition. This is now completely removed. Light clearing is likely fine (states call `startAnimation` which replaces the old one), but the serial flush loss means stale bytes could leak between states.

Suggest either:
- Add `flushSerial()` to `StateMachine::commitState()`, or
- Add it to individual `onStateDismounted()` methods that use serial (Idle, handshake states)

**3. Double map lookup in `Device::loop()`**

```cpp
// Current — find() result discarded, then operator[] re-traverses:
if(!appConfig.empty() && appConfig.find(currentAppId) != appConfig.end()) {
    appConfig[currentAppId]->onStateLoop(this);
}

// Cleaner:
auto it = appConfig.find(currentAppId);
if (it != appConfig.end()) {
    it->second->onStateLoop(this);
}
```

Trivial with 2-3 entries, but cleaner code.

---

### Performance Summary (ESP32-S3)

The perf team gave this a thorough review. TL;DR — no concerns:

| Category | Risk | Notes |
|---|---|---|
| Memory | LOW | ~2-4 KB per game app. Snapshots use `unique_ptr`, clean ownership. "Giant memory leak" is ~6-12 KB total — <1% of SRAM. |
| CPU | LOW | O(log 3) map lookup per frame. 1 extra virtual dispatch (~2 cycles). |
| Stack | LOW | ~300-500 bytes worst case during app switch. |
| Resource Safety | MEDIUM | `onStateChange()` removal (see #2 above). |
| Concurrency | LOW | No new issues. Buttons are synchronous. ESP-NOW race is pre-existing. |

---

### Nice to Have (future)

- Add `StateId::operator==` alongside the existing `operator<`
- Consider `StateId(-1)` sentinel for "no app" instead of defaulting to 0
- Guard `populateStateMap()` against double-call (if a non-resumed app gets `onStateMounted` twice, states would double-allocate)

---

### Compatibility

All APIs our FDN system depends on (`setActiveApp`, `AppConfig`, `Snapshot`, `onStatePaused/Resumed`, `StateMachine(int)`) are present and correctly implemented. We add `returnToPreviousApp()` and `getApp()` as clean extensions in PR #83 — no conflicts.

Nice work on the refactor! 🎯

PR #90, opened 2026-02-14
Break out player registration and handshake from quickdraw.

cyon1c:
* Create PlayerRegistrationApp and HandshakeApp
* Quickdraw now uses these two apps.
* Add match upload automatically in Player registration.
* Removed AllegiancePickerState, is unnecessary for offline mode and only created confusion.


FinallyEve (approved):
LGTM!

PR #108, opened 2026-04-06
ESP-NOW team coordination with chain duel support

tire-fire:
The chain duel feature required some changes at layers below the game. 

The RDC handshake now exchanges game role (isHunter) alongside device type, so getPeerIsHunter is available to any game. This touched the handshake wireless manager, the serial MAC string format in both input/output idle states, and the connected state handler to accept late EXCHANGE_ID packets for role exchange. A new kTeamCommand packet type was added to the peer comms layer, with the handler registered in main.cpp rather than inside Quickdraw. The ESP-NOW driver gained a destination MAC filter to drop stray broadcast packets not addressed to the local device, and a send queue drain on deinit to prevent leaked allocations.

### DIFF
--- a/include/game/quickdraw-states.hpp
+++ b/include/game/quickdraw-states.hpp
@@ -10,11 +10,16 @@
 #include "device/drivers/http-client-interface.hpp"
 #include "game/quickdraw-resources.hpp"
 #include <cstdlib>
+#include <functional>
 #include <queue>
 #include <string>
 #include "device/remote-device-coordinator.hpp"
+#include "wireless/team-packet.hpp"
 
-enum QuickdrawStateId {    
+class Quickdraw;
+#include "device/drivers/peer-comms-interface.hpp"
+
+enum QuickdrawStateId {
     SLEEP = 6,
     AWAKEN_SEQUENCE = 7,
     IDLE = 8,
@@ -25,7 +30,8 @@ enum QuickdrawStateId {
     DUEL_RESULT = 17,
     WIN = 18,
     LOSE = 19,
-    UPLOAD_MATCHES = 20
+    UPLOAD_MATCHES = 20,
+    SUPPORTER_READY = 21
 };
 
 class Sleep : public State {
@@ -69,7 +75,9 @@ private:
 
 class Idle : public ConnectState {
 public:
-    Idle(Player *player, MatchManager* matchManager, RemoteDeviceCoordinator* remoteDeviceCoordinator);
+    Idle(Player *player, MatchManager* matchManager, RemoteDeviceCoordinator* remoteDeviceCoordinator,
+         std::function<int()> getSupporterCount, std::function<void()> clearChainState,
+         std::function<bool()> isInviteRejected);
     ~Idle();
 
     void onStateMounted(Device *PDN) override;
@@ -81,10 +89,12 @@ public:
 private:
     Player *player;
     MatchManager* matchManager;
+    std::function<int()> getSupporterCount_;
+    std::function<void()> clearChainState_;
+    std::function<bool()> isInviteRejected_;
     bool matchInitialized = false;
     bool displayIsDirty = false;
     int statsIndex = 0;
-    int statsCount = 5;
 
     bool isPrimaryRequired() override;
     bool isAuxRequired() override;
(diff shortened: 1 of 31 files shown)


include/game/quickdraw-states.hpp
```diff
@@ -69,7 +74,7 @@ class AwakenSequence : public State {
 
 class Idle : public ConnectState {
 public:
-    Idle(Player *player, MatchManager* matchManager, RemoteDeviceCoordinator* remoteDeviceCoordinator);
+    Idle(Player *player, MatchManager* matchManager, RemoteDeviceCoordinator* remoteDeviceCoordinator, Quickdraw* quickdraw);
```
cyon1c: Rather than pass the whole quickdraw object, you could pass a function from quickdraw to the Idle state instead. Look at how std::bind is used for example in populateStatemap to pass the transition functions as function arguments.

include/game/quickdraw-states.hpp
```diff
@@ -305,3 +312,45 @@ class UploadMatchesState : public State {
     bool transitionToSleepState = false;
     bool shouldRetryUpload = false;
 };
+
+class SupporterReady : public State {
```
cyon1c: This should be extending ConnectState rather than just State.

src/apps/handshake/handshake-connected-state.cpp
```diff
@@ -53,6 +54,13 @@ void HandshakeConnectedState::heartbeatMonitorStringCallback(const std::string&
 void HandshakeConnectedState::listenForNotifyDisconnectCommand(HandshakeCommand command) {
     if (command.command == HSCommand::NOTIFY_DISCONNECT) {
         transitionToIdleState = true;
+    } else if (command.command == HSCommand::EXCHANGE_ID) {
```
cyon1c, line 57: this looks fishy to me, why are we entering notifyDisconnect and then not disconnecting here?

src/game/quickdraw-states/idle-state.cpp
```diff
@@ -52,19 +55,50 @@ void Idle::onStateMounted(Device *PDN) {
 }
 
 void Idle::onStateLoop(Device *PDN) {
+    int supporters = quickdraw_ ? quickdraw_->getSupporterCount() : 0;
```
cyon1c: Reiterating my above comment, you should be able to pass getSupporterCount as a function argument to the state rather than all of quickdraw.

src/game/quickdraw.cpp
```diff
@@ -25,7 +26,122 @@ Quickdraw::~Quickdraw() {
...
+            break;
+        case TeamCommandType::GAME_EVENT:
+            if (position > static_cast<int>(GameEventType::LOSS)) break;
+            if (memcmp(src, inviteChampionMac_, 6) != 0) break;
+            if (gameEventCallback_) gameEventCallback_(static_cast<GameEventType>(position));
+            break;
+    }
+}
+
+void Quickdraw::onStateLoop(Device* PDN) {
```
cyon1c, line 104: Why was this method added? onStateLoop in the base state-machine should handle all this already.

src/main.cpp
```diff
@@ -92,6 +92,14 @@ void setup() {
     Serial.begin(115200);
     while (!Serial) delay(100);
 
+    esp_reset_reason_t reason = esp_reset_reason();
```
cyon1c: why?

src/main.cpp
```diff
@@ -155,7 +163,13 @@ void setup() {
     setupEspNow(quickdrawWirelessManager, remoteDebugManager, peerCommsDriver);
     
     game = new Quickdraw(player, pdn, quickdrawWirelessManager, remoteDebugManager);
-    
+
+    // Register team packet handler alongside other ESP-NOW handlers
+    peerCommsDriver->setPacketHandler(PktType::kTeamCommand,
```
cyon1c: could this not be initialized in quickdraw constructor or initialize?

src/apps/handshake/handshake-connected-state.cpp
cyon1c: Second review of this and i'm seing it's renamed. nevermind.

src/apps/handshake/input-send-id-state.cpp
```diff
@@ -27,7 +28,13 @@ void InputSendIdState::onHandshakeCommandReceived(HandshakeCommand command) {
     }
 }
 
-void InputSendIdState::onStateLoop(Device *PDN) {}
+void InputSendIdState::onStateLoop(Device *PDN) {
```
cyon1c, line 31: how often are you seeing dropped packets that you felt this was necessary?

PR #110, opened 2026-04-14
RDC chain duel with posse display and loop detection

tire-fire:
Adds chain duel. Same-role chain peers boost champion's reaction time by 15ms each.

New: ChainDuelManager, SupporterReady state, 4 packet types (kChainAnnouncement/Ack/GameEvent/Confirm)

Changed in RDC: role byte added to announcement wire format, handshake delayed until game role is set so first announcement carries correct role, backward emit fixed to send other-side daisy chain (was echoing destination's own chain back causing pollution in 3+ chains), peer role only cleared on removal not add

Changed in game layer: MatchManager boost via callback instead of direct field, reaction time recorded at button press before boost subtraction (match stores boosted times so both sides agree on winner), DuelCountdown/Duel/Win/Lose broadcast game events via ChainDuelManager, Idle shows posse count with auto-refresh

Fixed: DuelResult::onStateDismounted cleared QWM packetReceivedCallback which was never re-registered, breaking back-to-back duels


tire-fire:
I'll close #108 if this is accepted instead.

cyon1c (approved):

Issue #111, opened 2026-04-21
WiFi task races against main loop


tire-fire:
ESP-NOW recv callbacks run on the WiFi task, and those callbacks write to MatchManager state while the main loop is reading the same fields causing a race. I put together a TSan test at commit `6a94fe8` and TSan flags races on every MatchManager getter the duel states read.

## TSan output

```
3x data race at src/game/match-manager.cpp:416   MatchManager::isMatchReady()
1x data race at src/game/match-manager.cpp:143   MatchManager::getHasPressedButton()
1x data race at src/game/match-manager.cpp:139   MatchManager::getHasReceivedDrawResult()
1x data race at include/game/match.hpp:91        Match::getHunterDrawTime()
1x data race at include/game/match.hpp:92        Match::getBountyDrawTime()
```

## Scope

I think this problem exists everywhere ESP-NOW packet handlers get registered:

```
src/main.cpp             : setPacketHandler(kQuickdrawCommand, ...)
src/main.cpp             : setPacketHandler(kDebugPacket, ...)
src/device/remote-device-coordinator.cpp : three setEspNowPacketHandler calls
src/game/quickdraw.cpp   : several setEspNowPacketHandler calls
```

Each one registers a callback on the WiFi task that touches state the main loop also reads.

## Possible fixes

1. Add a mutex to each manager. Works, reviewable in small chunks, but it's per-manager discipline, and anyone adding a new manager has to remember to do it.
2. Change the ESP-NOW driver so the recv callback enqueues the packet onto a thread-safe queue, and the main loop drains it. Handlers then run on the main task and there's nothing for them to race with. Makes wireless access deterministic but slower.

## Reproduce it

Check out the repo at `6a94fe8ea2de4dbcaf866fb632a9a4f85708ebfe`, copy your `wifi_credentials.ini` across, drop in the three pieces below, and run the test under TSan.

```bash
git worktree add /tmp/pdn-witness 6a94fe8ea2de4dbcaf866fb632a9a4f85708ebfe
cp wifi_credentials.ini /tmp/pdn-witness/
cd /tmp/pdn-witness
pio test -e native_tsan -f test_core --filter "*MatchManagerConcurrent*"
.pio/build/native_tsan/program --gtest_filter="*MatchManagerConcurrent*""
```

New file at `test/test_core/match-manager-concurrent.hpp`:

```cpp
#pragma once

#include <gtest/gtest.h>
#include <atomic>
#include <chrono>
#include <thread>
[...]

PR #112, opened 2026-04-21
Create a thread safe drain queue for processing incoming ESP-NOW packets.

cyon1c:
Closes Issue #111 .

ESP-NOW explicitly runs on Core 0, while the core application runs on Core 1. This results in possible race conditions.

The updated implementation adds a receiveQueue to the esp-now-driver, which is guarded with the recvMutex semaphore.

This does add a single loop iteration delay between packet receive and consumption, which shouldn't negatively impact game experience.


tire-fire:
LGTM

tire-fire:
Actually, now that I think about it, we should probably do the same protection against the send side as well.

PR #120, opened 2026-05-18
Peer-graph connectivity rewrite: binary framing, reliable transport, chain-manager

tire-fire:
Replaces the serial-string handshake and chain-roster loop detection with binary HELLO/BEACON serial framing, a peer-graph for loop detection, and a reliable ESP-NOW transport (Resender / ReliableChannel / WirelessTransport). ChainDuelManager becomes ChainManager. The duel-outcome model becomes set-once-per-side with void-on-abandon. Serial now carries only connectivity framing; game coordination is all over ESP-NOW.

Touches things you might not expect for a connectivity change:

- Match serialization: `Match::serialize`/`deserialize` and `MATCH_BINARY_SIZE` are removed; `toJson`/`fromJson` gained a `voided` field. Anything reading Match's binary form is gone, and the JSON shape changed.
- Player career stats: shootout matches no longer count toward wins/losses/streak/matchesPlayed (`countsTowardCareer` in `duel-result-state.cpp`).
- The ELIMINATED and ABORTED screens are one `ShootoutOutcome` class (two state IDs, value-parameterized). One deliberate delta: the ABORTED screen now clears button callbacks on mount; previously a duel aborted mid-flight kept a stale press handler live for the 2s screen.
- WiFi/radio: two excursion bugs fixed. An in-flight ESP-NOW send orphaned across `esp_now_deinit` left TX permanently dead after a WiFi excursion — `disconnect()` now clears the send queue. And the HTTP client arms `WiFi.setAutoReconnect(true)`; with no AP reachable the core then scans all channels every ~2.5s indefinitely, jamming ESP-NOW for every nearby device. The upload state now restores ESP-NOW mode on dismount, and the HTTP client disarms auto-reconnect in `disconnect()`. The HTTP client's connection flow is a single non-blocking path driven from `exec()` (the old `connect()` blocked the main loop for 5s polling `WiFi.status()`); a failed excursion gives up at 12.5s and parks the radio on the ESP-NOW fallback channel. `setWifiConfig`/`retryConnection`/`updateConfig` had no callers and are gone. Promiscuous mode is gated behind `PDN_ENABLE_RSSI_TRACKING`.
- Pin bias: the OUTPUT-jack RX line (GPIO 38) is now driven as a pulldown, not a pullup.

Watch out for:

- The per-match draw slot is not the player's allegiance. Shootout computes each side's slot from MAC ordering; duel mechanics read `MatchManager::localIsHunterForMatch()`, while chain and visuals read `Player::isHunter`.
- `onStateLoop` runs before `checkStateTransitions`, so a timer-driven teardown can act on a same-tick packet or press. `DuelPushed` guards with `!matchResultsAreIn()` before voiding; `DuelReceivedResult` guards `sendNeverPressed` with `!neverPressedSent_`. New states need the same guard.
- Match void is one-sided: if our grace expires before the opponent's reliable reply arrives, we void while they may keep a real result (comment in `duel-pushed-state.cpp`). Server `match_id` dedup absorbs it; local stats can drift by one.
- Supporter boost depends on the 1 Hz confirm backstop in `ChainManager::sync()`; without the periodic re-send a dropped confirm stops being counted.
- Chain boost is gated on `!currentMatchIsShootout()` at press time: closing a chain into a ring without unplugging leaves stale `confirmedSupporters` on ex-champions, and an unguarded provider read handed a tournament duelist a 30ms head start (decided a match on the bench). Don't replace the gate with confirm-clearing — that's timing-dependent.
- RDC's receive queue sheds oldest at 128 entries. Drop-oldest is only safe because every packet kind regenerates periodically (HELLO each emit tick, BEACON via the 1 Hz backstop, JACK_SILENT every watchdog pass) — if a non-regenerating kind is ever enqueued, this becomes silent loss.
- The 1 Hz backstop replays cached beacons of current *chain members only*. Replaying `allBeacons()` instead would resurrect departed nodes' beacons (zombie topology); keep the member filter.
- Role rides the BEACON as an opaque byte the topology layer floods and dedups but never interprets; the game layer reads it as tri-state (`chain_role::Role`: 0=NONE, 1=HUNTER, 2=BOUNTY — note 0 used to mean bounty). A NONE node (FDN/non-duelist, or anything that never installs the role provider) severs the duel chain in the champion walk instead of being misread as a bounty. The byte stays in `BeaconRecord::operator==` deliberately: a role flip must re-flood atomically with topology.
- The platform loop (`main.cpp`/`cli-main.cpp`) owns `WirelessTransport::sync()`; game managers must not pump it. QuickdrawWirelessManager is gone entirely — MatchManager owns the `kQuickdrawCommand` reliable channel directly, including `cancel()` semantics (cancel must NOT fire the abandon callback, or a stale abandon voids the next match against the same peer).

The two places most worth real scrutiny are the peer-graph mutual-edge and stability logic (loop-detection correctness) and the transport ack/dedup/retransmit path.

The why and the packet lifecycles live in the owning headers: `remote-device-coordinator.hpp` (serial lifecycle, backstops), `wireless-transport.hpp` (reliable ESP-NOW lifecycle), `wireless-manager.hpp` (radio exclusivity), `peer-graph.hpp` (derive-vs-maintain), `chain-manager.hpp` (role semantics).


lib/core/include/device/drivers/peer-comms-types.hpp
```diff
@@ -8,21 +9,34 @@ enum class PktType : uint8_t
...
+    kChainGameEvent = 3,
+    kChainConfirm = 4,
+    kShootoutCommand = 5,
+    kSymbolMatchCommand = 6,
+    kAck = 7,
     kNumPacketTypes //Not a real packet type, DO NOT USE
 };
 
+// Max length of a player name on the wire (no null terminator).
+inline constexpr size_t kNameLength = 12;
```
cyon1c, line 21: I know we have players with longer names than this - is there a technical limitation for it being 12? @tire-fire

lib/core/include/device/drivers/peer-comms-types.hpp
cyon1c, line 1: Is there a way to separate the game related logic out of core/ and into a more appropriate location?

lib/core/include/device/peer-graph-codec.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+// Payload lengths (bytes after the opcode, before the CRC). The single source of
+// truth for both the encoder here and SerialFrameParser's length checks.
+constexpr size_t kHelloPayloadLen = 9;   // source mac[6] + deviceType[1] + userId[2]
+constexpr size_t kBeaconPayloadLen = 19;  // source[6] + inPeer[6] + outPeer[6] + role[1]
+
+// 0xFFFF: peer has no registered 4-digit id yet (pre-pairing placeholder).
+constexpr uint16_t kUserIdUnknown = 0xFFFF;
+
+struct HelloRecord {
```
cyon1c, line 43: Is there a reason HELLO is defined here but not BEACON?

lib/core/include/device/peer-graph.hpp
```diff
@@ -0,0 +1,124 @@
...
+// peer MAC means "no peer on that jack". Equality is by full content so the
+// flood layer can dedup re-received beacons without sequence numbers.
+struct BeaconRecord {
+    net::Mac source{};
+    net::Mac inPeer{};
+    net::Mac outPeer{};
+    // Opaque game-layer byte (encoding: chain_role::Role); this layer floods
+    // and dedups it, never interprets it. Two hardware-tested invariants: it
+    // stays in operator== (a change must re-flood), and it rides the topology
+    // packet; a separate role flood diverged (all-supporters limbo).
```
cyon1c, line 19: This comment is incredibly opaque and confusing.

lib/core/include/device/mac-types.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+#include <array>
+#include <cstdint>
+
+// Shared 6-byte MAC address type and peer-validity helpers, used by the
+// peer-graph protocol, the RDC, and their tests.
+namespace net {
```
cyon1c, line 8: This style of namespace declaration seems fine, but i'd like to sort out how to make them a little more clear. Why are the mac types in a namespace named "net"? also, the `peer_graph` namespace is confusing as it can be conflated with the type.

lib/core/include/device/remote-device-coordinator.hpp
```diff
@@ -1,153 +1,383 @@
...
+    }
+
+    // Game-layer interpreters over the peer graph, injected by ChainManager:
+    // RDC owns the graph but not the meaning of its role byte. The virtual
+    // getters below stay the single read path so tests keep one seam to stub.
+    using ChampionResolver =
+        std::function<std::optional<net::Mac>(const PeerGraph&)>;
+    using OpponentRoleResolver =
+        std::function<std::optional<bool>(const PeerGraph&)>;
+    void setChampionResolver(ChampionResolver resolver) {
```
cyon1c, line 108: Is there a better way to approach this than having it in RDC? Couldn't we just have RDC provide a peer graph and then allow the games to do whatever they want with that peer graph?

lib/core/include/device/remote-device-coordinator.hpp
```diff
@@ -1,153 +1,383 @@
...
+    // player registration, the userId sense of "registered" above).
+    // unregisterRadioPeer also clears any direct-peer slot still naming the
+    // MAC, a no-op on the internal disconnect path (already gated on
+    // !isDirectPeer) but load-bearing for direct callers.
+    void registerRadioPeer(const uint8_t* macAddress);
+    void unregisterRadioPeer(const uint8_t* macAddress);
 
-    void setChainChangeCallback(std::function<void()> callback);
+    // Connected component of self under the peer-graph's mutual edges. Order is
+    // implementation-defined (callers must treat as a set). Virtual so test
```
cyon1c, line 135: Surely we can modify this so that we maintain an ordered vector of the chain utilizing our peer graph?

lib/core/include/device/remote-device-coordinator.hpp
```diff
@@ -1,153 +1,383 @@
...
+    // fixtures can subclass RDC and inject deterministic snapshots.
+    virtual std::vector<net::Mac> getChainMembers() const { return peerGraph_.getChainMembers(); }
 
-    // Direct peer drops only; daisy-chained drops arrive via chain announcements.
-    void setPeerLostCallback(std::function<void(const uint8_t*)> callback);
+    // True iff self is in a cycle: either the peer-graph finds a ring of mutual
```
cyon1c, line 139: what's the deal with the double f if all over this PR? is it short hand for "if and only if"?

lib/core/include/device/remote-device-coordinator.hpp
```diff
@@ -1,153 +1,383 @@
...
+    virtual std::optional<net::Mac> getChampionMac() const {
+        return championResolver_ ? championResolver_(peerGraph_) : std::nullopt;
+    }
 
-    using AnnouncementEmitCallback = std::function<void(const uint8_t* toMac, uint8_t announcementId, const std::vector<std::array<uint8_t, 6>>& peers)>;
-    void setAnnouncementEmitCallback(AnnouncementEmitCallback callback);
+    // Hunter/bounty role of self's opponent-jack peer when it forms a mutual
+    // edge, via the injected resolver (see chain_role::mutualOpponentIsHunter).
+    // Virtual for test override.
+    virtual std::optional<bool> mutualOpponentIsHunter() const {
```
cyon1c, line 155: This code smells. We will need to support other roles eventually, and it's more game logic in core.

lib/core/include/device/remote-device-coordinator.hpp
```diff
@@ -1,153 +1,383 @@
...
+                                     : std::nullopt;
+    }
 
-    void processChainAnnouncementAckPacket(const uint8_t* fromMac, const uint8_t* data, size_t dataLen);
+    // Run serviceConnectivity (HELLO emit + silent-link watchdog) on a fixed
+    // cadence; call this from a hardware timer/task. While set, sync() stops
+    // running serviceConnectivity itself (the task owns it). Native leaves it
+    // off, so sync() drives everything single-threaded for the tests.
+    void serviceConnectivity(unsigned long now);
+    void setExternalConnectivityTask(bool on) { externalConnectivityTask_ = on; }
```
cyon1c, line 165: I don't love the naming convention for these two functions. "serviceConnectivity" especially is an ambiguous.

lib/core/include/device/remote-device-coordinator.hpp
```diff
@@ -1,153 +1,383 @@
...
+    void declareJackDeadForTest(SerialIdentifier jack) {
+        declareJackDead(jack);
+    }
+
+    // Test hook to override the silent-link jack-dead threshold (production:
+    // kHelloSilentLinkMs). A single-device unit test has no partner emitting
+    // HELLOs, so the production threshold trips during any multi-second time
+    // advance (e.g. primeTopologyStableAll's 8.8s prime window). Tests that prime
+    // then exercise long-running behavior call this with a longer threshold.
+    void setJackDeadSilentLinkMsForTest(unsigned long ms) {
```
cyon1c, line 203: Confuuuuusing comment here.

lib/core/include/device/remote-device-coordinator.hpp
```diff
@@ -1,153 +1,383 @@
...
+    // 20ms (connectivityTask in main.cpp) and RX is timestamped on the UART event
+    // task within ~1ms of arrival, so the worst-case inter-HELLO gap is ~one emit
+    // period plus scheduling slack. 100ms is 5 missed emits: fast enough to feel
+    // instant, with margin against transient task slack. Overridable via
+    // setJackDeadSilentLinkMsForTest.
+    //
+    // Liveness is protocol-level (frames observed) rather than electrical
+    // (jack-sense GPIO) on purpose: the only question that matters is whether
+    // frames can flow, and TRS contacts can be electrically mated while passing
+    // nothing (partial insertion, oxidation). A GPIO-based disconnect detector
```
cyon1c, line 246: Some of these comments, including this one, seem to be telling a story about the work over the course of this PR rather than documenting the actual what/why/how of the code.

lib/core/include/device/remote-device-coordinator.hpp
```diff
@@ -1,153 +1,383 @@
...
-     * handshake send id state - connecting
-     * handshake connected state - connected
-     */
-    PortStatus mapHandshakeStateToStatus(SerialIdentifier port);
+    // Single writeBytes site for binary frames: increments framesTx.
+    void emitFrame(SerialIdentifier jack, const std::vector<uint8_t>& frame);
+
+    uint8_t portIndex(SerialIdentifier port) const;
+
+    static SerialIdentifier oppositeJack(SerialIdentifier j) {
```
cyon1c, line 274: why do we need this? When the FDN is merged in this code will become a problem, as an FDN has `INPUT_JACK` and `INPUT_JACK_SECONDARY`.

lib/core/include/device/remote-device-coordinator.hpp
```diff
@@ -1,153 +1,383 @@
...
+    SerialFrameParser frameParsers_[2];
+
+    PeerGraph peerGraph_;
+
+    // A received serial frame, parsed but not yet applied. Same idiom (and
+    // name) as EspNowDriver: ingestSerial only validates + enqueues; the main
+    // loop drains via exec() and runs the handlers single-threaded, so the
+    // peer-graph needs no lock.
+    struct DeferredPacket {
+        enum Kind { HELLO, BEACON, JACK_SILENT } kind;
```
cyon1c, line 302: First time seeing this `JACK_SILENT` enum. What is it's purpose? A comment here could be helpful explaining that JACK_SILENT is a synthetic event derived from a signal on device (when a jack was connected and stops receiving hello (I think???)).

lib/core/include/device/serial-frame-parser.hpp
```diff
@@ -0,0 +1,73 @@
...
+#include <cstddef>
+#include <cstdint>
+#include <functional>
+#include <vector>
+
+// A validated binary frame surfaced to downstream consumers (RDC's ingestSerial,
+// bound via setBinaryFrameHandler). Payload layouts are owned by
+// peer-graph-codec.hpp (kHelloPayloadLen / kBeaconPayloadLen); payload holds the
+// raw bytes between the opcode and the CRC.
+struct Frame {
```
cyon1c, line 13: SerialFrame as a slightly better name for this?

lib/core/include/device/serial-frame-parser.hpp
```diff
@@ -0,0 +1,73 @@
...
+    // Request a parser reset from another thread. The actual clear happens at
+    // the start of the next feed(), keeping parser state single-owner.
+    void requestReset() { resetRequested_.store(true); }
+
+    void setBinaryFrameHandler(BinaryFrameHandler cb) { binaryFrameHandler_ = std::move(cb); }
+
+private:
+    enum class ParseState {
+        ScanForSync,
+        GotAA,
```
cyon1c, line 41: Could be named more clearly, ie GotPreambleByte

lib/core/include/wireless/direct-peer-table.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include "mac-functions.hpp"
+#include "device/wireless-manager.hpp"
+#include "device/serial-manager.hpp"
+#include "device/device-type.hpp"
+
+struct DirectPeer {
+    std::array<uint8_t, 6> macAddr;
+    SerialIdentifier sid;
+    DeviceType deviceType;
+    uint16_t userId = 0xFFFF;  // 4-digit player code; 0xFFFF = not yet registered
```
cyon1c, line 22: Will FDN Support cause issues here?

lib/core/include/wireless/reliable-channel.hpp
```diff
@@ -0,0 +1,169 @@
...
+    PktType type() const { return type_; }
+    uint8_t subType() const { return subType_; }
+
+    void onAck(uint8_t seqId, const uint8_t* fromMac);
+    void onResenderAbandon(uint8_t seqId, const uint8_t* targetMac);
+
+    bool isPending(const uint8_t* mac) const;
+    void cancel(const uint8_t* mac);
+
+    // pendingCount counts peers with at least one in-flight entry on this channel.
```
cyon1c, line 38: comment is very unclear.

lib/core/include/wireless/reliable-channel.hpp
```diff
@@ -0,0 +1,169 @@
...
+    size_t pendingCount(const std::vector<std::array<uint8_t, 6>>& peers) const {
+        size_t n = 0;
+        for (const auto& p : peers) if (isPending(p.data())) ++n;
```
cyon1c, line 41: I have a slight preference against using auto except in very specific circumstances - it has a tendency to obfuscate meaning in code.

lib/core/include/wireless/reliable-channel.hpp
```diff
@@ -0,0 +1,169 @@
...
+protected:
+    uint8_t nextSeqId();
+    // Helper for the template subclass; defined in reliable-channel.cpp where
+    // WirelessTransport is complete.
+    WirelessManager* getWirelessManager() const;
+    // Own MAC, for self-exclusion in multi-target sends; nullptr without a transport.
+    const uint8_t* selfMac() const;
+    void sendOnceBytes(const uint8_t* mac, const uint8_t* data, size_t len);
+    // Defined in the .cpp so the logger stays out of this template header.
+    static void logLengthMismatch(PktType type, uint8_t subType,
```
cyon1c, line 61: unclear.

lib/core/include/wireless/reliable-channel.hpp
```diff
@@ -0,0 +1,169 @@
...
+        uint8_t lastSeqId;
+    };
+    OnAbandon onAbandon_;
+    uint8_t lastSeqId_ = 0;
+    // Last nonzero seqId delivered per sender. Bounded by the physical peer
+    // count present in a session.
+    std::vector<RxSeqRecord> rxSeq_;
+};
+
+template <class P, class Sub = void>
```
cyon1c, line 89: what is going on with this template?

lib/core/include/wireless/reliable-channel.hpp
```diff
@@ -0,0 +1,169 @@
...
+    // On a sub-typed channel the payload's leading cmd byte is the receiver's
+    // demux key; stamping it here (like seqId) means a send site can never
+    // route its payload to a sibling channel by mis-stating the cmd.
+    void stampSubType(P& p) {
+        if constexpr (!std::is_same_v<Sub, void>) {
+            p.cmd = subType_;
+        }
+    }
+
+    template <typename SendToTarget>
```
cyon1c, line 159: When using templates, we should have comments explaining what an appropriate type could be in the context.

lib/core/include/wireless/resender.hpp
```diff
@@ -0,0 +1,157 @@
...
+        : wm_(wirelessManager) {}
+    ~Resender() = default;
+
+    // Send a kAck for a reliably-received packet. All subsystems share
+    // this one ack wire format.
+    static void sendAck(WirelessManager* wm, const uint8_t* toMac,
+                        PktType originalType, uint8_t subType, uint8_t seqId);
+
+    void setAbandonCallback(AbandonCallback cb) {
+        abandonCallback_ = std::move(cb);
```
cyon1c, line 86: I've noticed std::move in several places in the code - Are we performing this operation on `unique_ptr` objects, or just plain ol' pointers? If on plain ol pointers, is there a reason we're using move instead of just a standard assignment op?

lib/core/src/device/remote-device-coordinator.cpp
```diff
@@ -1,383 +1,420 @@
...
-            retryStats_.ackLatencyMsSum += pending.timer.getElapsedTime();
-            retryStats_.ackCount++;
-            pending.active = false;
-            pending.timer.invalidate();
-            return;
+    for (SerialIdentifier jack :
+         {SerialIdentifier::INPUT_JACK, SerialIdentifier::OUTPUT_JACK}) {
+        SerialFrameParser& parser = frameParsers_[portIndex(jack)];
+        parser.setBinaryFrameHandler(
+            [this, jack](const Frame& f) { ingestSerial(jack, f); });
```
cyon1c, line 55: Strong prefer for std::bind style semantics for functions like this.

lib/core/include/device/peer-graph-codec.hpp
```diff
@@ -0,0 +1,56 @@
...
+//                  connect, like the MAC.
+//   BEACON (0x01): flooded topology frame, payload = source + inPeer + outPeer
+//                  + role[1], an opaque game byte carried verbatim (see
+//                  BeaconRecord::role for why it rides this flood).
+//
+// encode*() produces a full framed packet for transmission. decode*() is the
+// inverse over the *payload* a SerialFrameParser hands back once it has stripped
+// and CRC-checked the framing, so framing/CRC live in exactly one place (the
+// parser) and payload<->record in exactly one place (here).
+namespace peer_graph {
```
cyon1c, line 25: Why is this a namespace and not a base class?

After further inspection, I find it strange that the codec includes the peer-graph type, rather than the other way around? If there is clarification for why this way, that's fine, but encapsulation logic could probably be a little more coherent here.

lib/core/src/device/remote-device-coordinator.cpp
```diff
@@ -1,383 +1,420 @@
...
 
-PortStatus RemoteDeviceCoordinator::getPortStatus(SerialIdentifier port) {
-    PortStatus statusByState = mapHandshakeStateToStatus(port);
-
-    if (statusByState == PortStatus::CONNECTED && !daisyChainedByPort_[portIndex(port)].empty()) {
-        return PortStatus::DAISY_CHAINED;
-    }
-
-    return statusByState;
+void RemoteDeviceCoordinator::emitFrame(SerialIdentifier jack,
```
cyon1c, line 174: emit is confusing language in this context - we have lots of examples of send/receive, no need to add a second word for "send", either that or a more descriptive name.

lib/core/src/device/remote-device-coordinator.cpp
```diff
@@ -1,383 +1,420 @@
...
+    if (!serialManager) return;
+    serialManager->writeBytes(frame.data(), frame.size(), jack);
+    stats_[portIndex(jack)].framesTx++;
 }
 
-PortState RemoteDeviceCoordinator::getPortState(SerialIdentifier port) {
-    std::vector<std::array<uint8_t, 6>> peerAddresses;
-
-    const Peer* macPeer = handshakeWirelessManager.getMacPeer(port);
+unsigned long RemoteDeviceCoordinator::nowMs() const {
```
cyon1c, line 181: we could probably add a global static method for this, as its probably useful to reuse.

lib/core/src/device/remote-device-coordinator.cpp
```diff
@@ -1,383 +1,420 @@
...
-                ? SerialIdentifier::OUTPUT_JACK : SerialIdentifier::INPUT_JACK;
-            const Peer* otherDirect = handshakeWirelessManager.getMacPeer(otherPort);
-            if (otherDirect != nullptr) {
-                // Forward: tell other side that nothing is reachable through changed_port
-                emitAnnouncementVia(otherPort, {});
+    const unsigned long now = nowMs();
+    while (!batch.empty()) {
+        DeferredPacket ev = batch.front();
+        batch.pop();
+        if (ev.kind == DeferredPacket::HELLO) {
```
cyon1c, line 127: prefer switch statement here for a well typed if/else if/else statement.

lib/core/src/device/remote-device-coordinator.cpp
```diff
@@ -1,383 +1,420 @@
...
+
+    // Connectivity producer: emit HELLO + run the watchdog. On hardware a 20ms
+    // task owns this (externalConnectivityTask_ set), so sync() skips it; on
+    // native it runs inline so the tests drive identical logic via sync().
+    if (!externalConnectivityTask_) serviceConnectivity(now);
+
+    // Drain the receive queue: RX HELLO/BEACON (macPeer, accept+flood) and the
+    // watchdog's JACK_SILENT (declareJackDead), all single-threaded, before
+    // the direct-peer scan and reconcile read macPeer.
+    exec();
```
cyon1c, line 257: I got very confused why the RDC has both exec and sync. exec could probably be renamed to something like "drainReceiveQueue" or something similar.

lib/core/src/device/remote-device-coordinator.cpp
```diff
@@ -1,383 +1,420 @@
...
+
+    // Runs last so connect/disconnect/jack-dead mutations from this tick are
+    // captured.
+    reconcileSelfPeers();
+
+    // The role byte is BEACON content like the peer slots: re-emit on a change
+    // rather than waiting for the backstop. emitBeaconBothJacks() is also what
+    // copies the byte into the local graph (setSelfRole), so without this the
+    // local champion walk reads a stale role for up to a backstop period.
+    const uint8_t roleNow = roleProvider_ ? roleProvider_() : 0;
```
cyon1c, line 270: I still don't fully understand why this is necessary.

lib/core/include/protocol-constants.hpp
```diff
@@ -1,17 +1,9 @@
 #pragma once
 
-#include <string>
+#include <cstdint>
 
-// Serial communication speed shared by all serial drivers and the handshake protocol.
+// Serial communication speed shared by all serial drivers.
 constexpr uint16_t BAUDRATE = 19200;
```
cyon1c, line 6: move into serial wrapper file, delete this file.

lib/core/include/wireless/wireless-transport.hpp
```diff
@@ -0,0 +1,115 @@
...
+//     (which itself retries a failed MAC-layer send). The platform loop's
+//     transport->sync() retransmits on the RetryPolicy backoff until the ack
+//     lands or retries exhaust -> the channel's abandon callback. Abandonment
+//     is a game-level signal (void a match, abort a tournament), not a log
+//     line; channel->cancel() drops pending sends WITHOUT it.
+//   receive: driver rx -> per-PktType handler (installed here: kAck at
+//     construction, data types when their first channel is claimed) ->
+//     deliverIncoming, which acks BEFORE dedup (a retransmit means our first
+//     ack was lost: re-ack, then drop the duplicate) and dispatches the
+//     decoded payload to the owning channel's onReceive. Claiming a channel
```
cyon1c, line 31: function naming conventions in this file could be improved.

Issue #169, opened 2026-07-01
game: refactor ShootoutManager — coordinator claim via RDC, ring-closed broadcast


cyon1c:
Move coordinator ownership into `ShootoutManager` and replace polling-based topology checks with RDC callbacks. `ChainManager` dependency is removed.

**Scope**
- Subscribe to `RDC::setOnRingClosed`: claim coordinator status when the callback fires. No election algorithm — the head at ring detection time is coordinator. The existing `BRACKET_ENTRY` lower-MAC stand-down handles the simultaneous-ring race and remains unchanged
- On ring closure: broadcast a ring-closed `ReliableChannel` packet to all roster members (`rdc_->getChainMembers()`). Members enter `ShootoutProposal` on receipt — replaces each device independent `isInStableLoop()` check, which is not available without BEACON
- `buildLoopMemberSet()` calls `rdc_->getChainMembers()` instead of `chainManager_->getSupporterChainPeers()`
- `isInStableLoop()` calls `rdc_->isInRing()`
- `isRingSettledOpen()` calls `rdc_->getChainRole() != Ring && rdc_->isTopologyStable()`
- Remove `ChainManager*` dependency from `ShootoutManager`
- Seven `ReliableChannel` registrations and `TournamentState` reset pattern are unchanged

**Blocked by:** #159

**Reference**
[game-layer-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/145) — ShootoutManager after redesign; [rdc-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/144) — `setOnRingClosed` as coordinator claim point

Issue #152, opened 2026-07-01
esp-now: hardware ACK as delivery signal; collapse dual retry layers; remove kAck packet type


cyon1c:
`ESP_NOW_SEND_SUCCESS` is a MAC-layer ACK — the peer radio confirmed receipt. Currently this signal is discarded and `Resender` waits for an application-layer `kAck` round-trip. Promote the hardware ACK to the delivery signal and collapse the dual retry architecture.

**Scope**
- In `EspNowSendCallback`: on `SEND_SUCCESS`, call `resender.onAck(packetId)` directly instead of waiting for a `kAck` packet
- On `SEND_FAIL` (all driver retries exhausted), trigger a `Resender` retry cycle directly instead of waiting for round-trip timeout
- Remove the `kAck` packet type, `Resender::sendAck()`, and the `kAck` recv handler
- Fix the shared-counter bug: the synchronous retry loop in `SendFrontPkt()` and the async callback in `EspNowSendCallback` share `m_curRetries`. Split into separate counters so synchronous enqueue failures do not consume async retry budget

**Blocked by:** #148

**Reference**
[esp-now-simplification-opportunities](https://github.com/alleycatassetacquisitions/pdn/discussions/143) — Findings 1 & 3

tire-fire:
Landed in #190 (merged to comms-rewrite).

Issue #151, opened 2026-07-01
transport: port WirelessTransport from PR; add sumStats() utility


cyon1c:
Bring `WirelessTransport` into the codebase and extend it with a `sumStats()` utility that replaces the per-manager aggregation boilerplate currently duplicated across game managers.

**Scope**
- Port `WirelessTransport` cleanly from PR, stripping any BEACON-related registration
- Add `Resender::Stats sumStats(std::initializer_list<PktType> types) const` — callers pass their own packet type list and receive a rolled-up retry view
- `ChainManager::getRetryStats()` switches to `sumStats()` in #168, or note as TODO

**Blocked by:** #148

**Reference**
[rdc-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/144) — WirelessTransport stats utility

tire-fire:
Covered by PR #175 together with #148 — the trio doesn't compile separately. Note: the sumStats() scope item was deliberately dropped (per-channel stats replaced by retransmit/abandon log events); see the PR description. Proposing this issue closes as folded into #148.

tire-fire:
Correction: this stays its own issue/PR after all. The compile coupling that suggested folding into #148 was one removable indirection (channels held WirelessTransport* only to fetch the WirelessManager; they now take WirelessManager* directly). WirelessTransport is implemented on branch tire-fire:151-wireless-transport, stacked on PR #175 — PR opens when #175 merges. Note the sumStats() scope item is still proposed dropped (per-channel stats were replaced by retransmit/abandon log events in #148; see PR #175).

Issue #148, opened 2026-07-01
transport: implement Resender and ReliableChannel


cyon1c:
Bring `Resender` and `ReliableChannel<P>` into the codebase. These are the foundational reliable-send primitives that RDC and the game layer build on. Port cleanly from the PR branch (`tire-fire/reliable-duel-delivery`), stripping any BEACON-specific hooks.

**Scope**
- `Resender`: application-layer retry with exponential backoff (100/200/400ms), sequence numbering, deduplication
- `ReliableChannel<P>`: typed wrapper over `Resender`; `sendReliable(P)`, `deliver(data, len)`, `onAck(id)`
- No BEACON-specific packet types or callbacks
- `kAck` packet type is addressed separately in #152

**Out of scope**
`WirelessTransport` (separate issue #151). Any game-specific packet types.

**Unblocks**
#151, #152

**Reference**
[rdc-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/144) — Architecture section

tire-fire:
Implemented in PR #175 (includes #151 — WirelessTransport/ReliableChannel/Resender are one compile unit; suggest folding #151 into this). Deviations from the source branch are listed in the PR description.

tire-fire:
Correction to my earlier comment: PR #175 is now scoped to #148 only (Resender + ReliableChannel). The transport-pointer indirection that coupled the channel code to WirelessTransport was removed (channels take WirelessManager* directly), so #148 and #151 build and review independently — no fold needed. #151 lands as its own PR (branch tire-fire:151-wireless-transport, stacked on this) once #175 merges. The two driver fixes moved to PR #176.

PR #175, opened 2026-07-03
transport: Resender and ReliableChannel (#148)

tire-fire:
Ports `Resender` and `ReliableChannel` from `tire-fire/reliable-duel-delivery` (the branch the 6-device hardware validation ran on) onto the clean base. Scope is #148 only; `WirelessTransport` follows as its own PR (#151) stacked on this one.

**Deviations from "port cleanly" — the parts worth reviewing hardest:**

- **Channels take `WirelessManager*` directly.** The source had channels hold a `WirelessTransport*` solely to call `getWirelessManager()` through it. That indirection was the only thing coupling #148's files to #151's, so it's removed — which is what makes the two issues independently buildable and reviewable. The transport passes its manager in at channel construction (visible in the #151 PR).
- **subType is removed.** Channel identity is the PktType alone; the ack payload is 2 bytes (`originalType`, `seqId`). Shootout's seven commands get their own PktTypes when #169 lands.
- **Per-channel stats are replaced by two log events** (retransmit `LOG_W`, abandon `LOG_E`, tag `RSND`). This also deletes #151's `sumStats` scope. If bench debugging misses the counters, this is the piece to argue about.
- **`kAck` is appended at 15**, not renumbered per the source — the legacy enum entries (incl. the four old per-manager `*Ack` types) stay untouched until their owners are deleted by later issues.

**Fragile spots:** `Resender::sync()` (abandon collection is deferred past iteration so callbacks can safely re-enter) and `ReliableChannel::deliver()` (ack-before-dedup ordering — a duplicate must still re-ack to silence the sender, but dispatch exactly once).

Adjacent-scope heads-up: #152's "hardware ACK as delivery signal; remove kAck" clause is unsound — `SEND_SUCCESS` is a MAC-layer ack that fires even when the receiver's app is deaf (mode transitions, our documented WiFi-excursion windows). Evidence and a proposed re-scope coming on #143/#152; this PR keeps kAck accordingly.

Verification: this commit standalone passes 267/267 native tests (4 new: seqId wrap, dedup-table eviction, keep-distinct slot independence, supersede + stale-ack). With the stacked #151 commit: 276/276.

### DIFF
--- a/lib/core/include/wireless/reliable-channel.hpp
+++ b/lib/core/include/wireless/reliable-channel.hpp
@@ -1,0 +1,179 @@
+#pragma once
+
+#include <array>
+#include <cstdint>
+#include <cstring>
+#include <functional>
+#include <vector>
+
+#include "device/drivers/peer-comms-types.hpp"
+#include "wireless/resender.hpp"
+
+// A ReliableChannel is a typed, reliable pipe for exactly one PktType: game
+// code sends a packed payload struct and receives decoded structs back, and
+// the channel supplies everything reliability needs underneath — seqId
+// stamping, retry-until-ack via the shared Resender, ack emission on receipt,
+// duplicate suppression, and an abandon callback when a peer stays
+// unreachable past the retry budget. The struct itself is the wire format
+// (packed, memcpy'd); both ends run the same firmware, so field layout is the
+// protocol. ReliableChannelBase holds the untyped mechanics; the
+// ReliableChannel<P> template below binds them to a payload type. Channels
+// are created and owned by WirelessTransport, one per PktType.
+class ReliableChannelBase {
+public:
+    using OnAbandon = std::function<void(uint8_t seqId, const uint8_t* targetMac)>;
+
+    /// wirelessManager may be nullptr in probe-style unit tests; every use is
+    /// null-tolerant.
+    ReliableChannelBase(WirelessManager* wirelessManager,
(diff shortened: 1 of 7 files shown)


lib/core/include/wireless/reliable-channel.hpp
```diff
@@ -0,0 +1,148 @@
...
+#include <array>
+#include <cstdint>
+#include <cstring>
+#include <functional>
+#include <vector>
+
+#include "device/drivers/peer-comms-types.hpp"
+#include "wireless/resender.hpp"
+
+class ReliableChannelBase {
```
cyon1c, line 22: A little document block here to give a baseline of what a ReliableChannel is would be helpful.

lib/core/src/wireless/reliable-channel.cpp
```diff
@@ -0,0 +1,86 @@
+#include "wireless/reliable-channel.hpp"
+#include "device/wireless-manager.hpp"
+#include "device/drivers/logger.hpp"
+
+namespace {
+constexpr const char* WTX_TAG = "WTX";
```
cyon1c: Feel like the tag should just be "ReliableChannel", right?

cyon1c (approved):
Code styling needs to be addressed - one character symbols, trailing underscores, very dense symbol names, but code is sound. Give this a pass with the code style guide and then it should be g2g.

Issue #147, opened 2026-07-01
esp-now: driver cleanup — RSSI via recv callback, remove promiscuous path, migrate deprecated send-callback param


cyon1c:
Three small, independent cleanup items in `EspNowDriver`, bundled because they all touch the send/receive callback surface.

**Scope**
- Read `esp_now_info->rx_ctrl->rssi` directly in `EspNowRecvCallback`. Delete `WifiPromiscuousRecvCallback`, the `PDN_ENABLE_RSSI_TRACKING` flag, and the `m_rssiTracker` fill path. No behavior change to RSSI consumers; only the data source changes
- Switch the send callback from the deprecated `status` parameter to `esp_now_info->tx_status` (field of `wifi_tx_info_t`). The old parameter is documented for removal in a future IDF release

**Out of scope**
Any change to how RSSI data is consumed upstream.

**Reference**
[esp-now-simplification-opportunities](https://github.com/alleycatassetacquisitions/pdn/discussions/143) — Findings 4 & 6

PR #176, opened 2026-07-03
esp-now: port hardware-proven driver fixes from fork

tire-fire:
Ports two hardware-proven ESP-NOW driver fixes from `tire-fire/reliable-duel-delivery` that the clean-base strategy lost. No board issue covers them (candidates for #147 or their own); flagging here so the rewrite doesn't ship with two regressions we already debugged once.

- **`clearSendQueue()` in `disconnect()`** — `esp_now_deinit()` destroys the TX-complete callback the send queue drains on, so an in-flight send at teardown orphans the queue permanently: ESP-NOW TX bricks after a WiFi excursion while RX keeps working. Diagnosed and validated on the bench (originally commit 2c26254).
- **Readback-verified channel pin + 100ms `exec()` retry** — `esp_wifi_set_channel` ESP_FAILs while a STA connect/scan is in flight, so trust the readback, not the return code, and re-pin from `exec()` until it sticks. Recovered in 106ms when the race re-fired on hardware. Includes `setAutoReconnect(false)` (the retry re-issues `disconnect()` to kill the blocking STA attempt, which is futile if auto-reconnect keeps rescheduling scans) and `PS_NONE` (modem-sleep with no AP drops L2 ACKs).

Both are verbatim ports of proven code; the only editorial change is splitting them into one commit per fix. Verification: esp32-s3 release firmware compiles; behavior was validated on hardware when the fixes originally landed on the fork.

### DIFF
--- a/lib/esp32-drivers/include/device/drivers/esp32-s3/esp-now-driver.hpp
+++ b/lib/esp32-drivers/include/device/drivers/esp32-s3/esp-now-driver.hpp
@@ -45,6 +45,21 @@ public:
     // === PEER COMMS INTERFACE === //
 
     void exec() override {
+        // Re-pin armed by connect(): retry (paced) until the readback sticks,
+        // re-issuing the disconnect each attempt to kill whatever STA attempt
+        // is blocking the pin. Stops if an excursion takes the radio.
+        if (channelPinPending_ && peerCommsState == PeerCommsState::CONNECTED) {
+            unsigned long now = millis();
+            if (now - lastChannelPinMs_ >= kChannelPinRetryMs) {
+                lastChannelPinMs_ = now;
+                WiFi.disconnect(false);
+                if (pinEspNowChannel()) {
+                    channelPinPending_ = false;
+                    LOG_W("ENC", "channel pin recovered: on %d", ESPNOW_CHANNEL);
+                }
+            }
+        }
+
         std::queue<DeferredPacket> pending;
         xSemaphoreTake(recvMutex_, portMAX_DELAY);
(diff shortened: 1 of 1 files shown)


lib/esp32-drivers/include/device/drivers/esp32-s3/esp-now-driver.hpp
```diff
@@ -45,6 +45,21 @@ class EspNowManager : public PeerCommsDriverInterface
     // === PEER COMMS INTERFACE === //
 
     void exec() override {
+        // Re-pin armed by connect(): retry (paced) until the readback sticks,
+        // re-issuing the disconnect each attempt to kill whatever STA attempt
+        // is blocking the pin. Stops if an excursion takes the radio.
+        if (channelPinPending_ && peerCommsState == PeerCommsState::CONNECTED) {
+            unsigned long now = millis();
+            if (now - lastChannelPinMs_ >= kChannelPinRetryMs) {
+                lastChannelPinMs_ = now;
```
cyon1c: Did you get the code style rules implemented @tire-fire ? I know this file doesn't follow it fully, variable names should be camelCase, no _.

cyon1c (approved):
Approving because only issue is code style clean up. everything else is g2g.

lib/esp32-drivers/include/device/drivers/esp32-s3/esp-now-driver.hpp
tire-fire: No I didn't, let me work on that.

cyon1c (approved):

Issue #161, opened 2026-07-01
rdc: implement edge case guards


cyon1c:
Five specific defensive behaviors called out in the design doc. Each is small but guards against a real failure mode observed in the prior implementation.

**Scope**
1. **Self-HELLO echo rejection** — reject HELLOs where `source == selfMac` _before_ updating `lastHelloRxMs_`. If liveness timestamp is updated first, a yanked output cable stays "alive" forever via the echo loop
2. **ESP-NOW deregistration guard** — only release an ESP-NOW peer slot when the dropping MAC is not present on the _other_ jack. A two-device ring has the same MAC on both jacks; removing the slot on one-cable disconnect silently breaks wireless for the still-connected device
3. **Watchdog underflow clamp** — if `lastHelloRxMs_ >= now` (concurrent task stamp race), treat gap as 0 rather than underflowing to `~UINT32_MAX` and firing a false dead-link
4. **Liveness baseline seed on connect** — seed `lastHelloRxMs_ = nowMs()` immediately on first peer detection, before first HELLO can arrive; prevents watchdog from firing in the connection window
5. **Serial frame parser reset on dead jack** — call `SerialFrameParser::requestReset()` on the dead jack when declaring it idle; mid-frame state from the dying connection corrupts the next device to plug in

**Blocked by:** #155

**Reference**
[rdc-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/144) — Edge case checklist, "Must be retained verbatim" section

tire-fire:
All five guards are in the merged comms-rewrite code; closing as complete. Mapping:
1. Self-HELLO echo rejection — remote-device-coordinator.cpp onHelloReceived: zero-source and self-MAC rejected before any liveness stamp (tested: rdcHelloRejectsSelfAndZeroSource).
2. ESP-NOW deregistration guard — releaseHelloPeer: other-jack scan + daisy-record guard keep the slot (tested: rdc2NodeRingSingleJackDropKeepsPeerSlot); #213 adds a held-head guard.
3. Watchdog underflow clamp — HelloLinkContext::sinceHello() clamps to 0 when now < lastHelloMs. Untested; test moves to #173.
4. Liveness baseline seed — obsolete by design: detection IS HELLO receipt, and onHelloReceived stamps lastHelloMs in the same call that opens the link, so no pre-HELLO watchdog window exists.
[...]

Issue #153, opened 2026-07-01
rdc: define HelloPayload wire format; integrate into SerialFrameParser


cyon1c:
Define the `HelloPayload` packed struct and implement encode/decode using `reinterpret_cast` + `memcpy` — the same pattern as `peer-comms-types.hpp`. Wire it into `SerialFrameParser` as the sole serial payload type. This is the concrete wire contract that all RDC work builds on. Also covers Finding 7 (serial payload encoding unification) — no separate issue needed.

**Scope**
- Define `HelloPayload` (14 bytes, packed): `source[6]`, `deviceType`, `headMac[6]`, `confirmed`
  - `source`: this device MAC
  - `deviceType`: PDN / FDN hardware fact; serves as discriminator for `ConnectionContext` deserialization
  - `headMac`: all-zero if head or standalone; propagated hop-by-hop for head conflict resolution and ring detection
  - `confirmed`: 1 when context exchange with head is complete; 0 while negotiating or after a head change not yet re-confirmed
- `encodeFramed` wraps a `HelloPayload` into a framed serial message
- `SerialFrameParser` fires a callback with a deserialized `HelloPayload` on each valid frame

**Blocked by:** #149

**Reference**
[rdc-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/144) — HELLO wire format section

Issue #149, opened 2026-07-01
transport: port SerialFrameParser from PR


cyon1c:
Bring `SerialFrameParser` and its matching `encodeFramed` utility into the codebase. This is the framing layer that gives UART message-boundary semantics — equivalent to what 802.11 gives ESP-NOW for free.

**Scope**
- `SerialFrameParser`: stateful parser that reconstructs framed messages from a raw byte stream
- `encodeFramed`: wraps a payload in a length-prefixed, CRC-checked frame
- Support HELLO opcode only; the BEACON opcode present in the PR is not ported
- `SerialFrameParser::requestReset()` must be present (used by RDC edge case guards in #161)

**Out of scope**
`HelloPayload` definition (separate issue #153). Any serial driver changes.

**Unblocks**
#153

**Reference**
[rdc-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/144) — HELLO wire format; [esp-now-simplification-opportunities](https://github.com/alleycatassetacquisitions/pdn/discussions/143) — Finding 7

PR #177, opened 2026-07-03
transport: SerialFrameParser and encodeFramed (#149)

tire-fire:
Ports `SerialFrameParser` (+ `crc16.hpp`, verbatim) from `tire-fire/reliable-duel-delivery` and extracts `encodeFramed` out of the fork's BEACON codec into the parser's own files. Scope is #149; independent of the transport PRs (#175).

**Deviations from "port cleanly" — worth reviewing hardest:**

- **HELLO is the sole opcode.** The fork's BEACON opcode and `peer-graph-codec` are not ported (rejected design). The opcode table collapses to `kOpHello`, and anything else — including the fork's old `0x01` BEACON — drops and resyncs (tested).
- **Framing constants re-homed.** Preamble bytes, `kOpHello`, and the HELLO payload length lived in the codec header; they now live with the parser, which is the other half of the same wire contract.
- **`kHelloPayloadLen = 14`** — the redesigned HELLO (source[6] + deviceType[1] + headMac[6] + confirmed[1]), not the fork's 9-byte userId variant. The packed `HelloPayload` struct itself is #153's scope; the parser only needs the length to frame it. If #153 changes the field set, this constant moves with it.
- **Two tests the fork didn't have:** an `encodeFramed`→parser round-trip (encoder and parser are cross-checked against an independent hand-built frame), and a `requestReset()` test — the cross-thread reset is required by #161's edge-case guards and was previously untested.

**Fragile spots:** the mid-frame timeout clamp in `checkTimeout()` (backwards-clock underflow guard — mirrors the same guard pattern the RDC rewrite specifies) and the GotAA-stays-on-0xAA resync subtlety in `feed()`.

Verification: 273/273 native tests (10 parser tests: framing, CRC rejection with resync, unknown-opcode drop incl. old BEACON, fragmentation, mid-frame timeout, cross-thread reset, garbage resync, double-preamble, split-preamble).

### DIFF
--- a/lib/core/include/device/serial-frame-parser.hpp
+++ b/lib/core/include/device/serial-frame-parser.hpp
@@ -1,0 +1,93 @@
+#pragma once
+
+#include <atomic>
+#include <cstddef>
+#include <cstdint>
+#include <functional>
+#include <vector>
+
+// Serial frame protocol constants. HELLO is the sole opcode on the wire:
+// serial carries physical-connectivity discovery/liveness only, everything
+// else rides ESP-NOW. The framing (preamble + opcode + payload + CRC-16)
+// recreates for UART what 802.11 gives ESP-NOW for free: message boundaries
+// and integrity on a raw byte stream with flaky TRS contacts.
+constexpr uint8_t FRAME_PREAMBLE_0 = 0xAA;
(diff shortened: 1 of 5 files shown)


cyon1c (approved):
Code style pass necessary, but looks good to me.

cyon1c (approved):

PR #179, opened 2026-07-03
esp-now: driver cleanup — RSSI via recv callback, drop promiscuous, tx_status (#147)

tire-fire:
Two independent driver cleanups from discussion #143 (Findings 4 & 6). No behavior change to RSSI consumers — only the data source moves.

**Finding 6 — RSSI without promiscuous mode.** `esp_now_recv_info_t::rx_ctrl->rssi` is present on every ESP-NOW receive, so the driver now fills `rssiTracker` directly in `EspNowRecvCallback`. Deleted `WifiPromiscuousRecvCallback`, the `esp_wifi_set_promiscuous*` setup in the constructor, and the `PDN_ENABLE_RSSI_TRACKING` flag (which was already dead — it gated nothing, only a misleading comment). This removes the UAP-bypass risk the promiscuous path carried (uncabled devices latching each other as wireless peers). The consumer surface (`getRssiForPeer` → `rssiTracker`, read by `remote-player-manager.cpp`) is untouched.

**Finding 4 — deprecated send-callback param.** `EspNowSendCallback` now reads `esp_now_info->tx_status` (field of `wifi_tx_info_t`) instead of the `status` parameter, which the IDF header documents for removal. The callback signature is fixed by `esp_now_register_send_cb`, so the `status` param stays for ABI compatibility but is no longer read. The two enums share values (`ESP_NOW_SEND_SUCCESS == WIFI_SEND_SUCCESS`), so the comparison is against `WIFI_SEND_SUCCESS`.

Net −23 lines. Verification: esp32-s3 release firmware compiles; 277/277 native tests (the RSSI consumer path in remote-player-manager stays green). Field names confirmed against the pinned toolchain header (IDF 5.5).

### DIFF
--- a/lib/esp32-drivers/include/device/drivers/esp32-s3/esp-now-driver.hpp
+++ b/lib/esp32-drivers/include/device/drivers/esp32-s3/esp-now-driver.hpp
@@ -17,13 +17,6 @@
 
 #define DEBUG_PRINT_ESP_NOW 0
 
-//Change to 1 to enable tracking rssi for peers
-//This works, but requires wifi to be in promiscuous mode
-//which likely prevents connecting to access points and
-//requires an unknown but likely high amount of processing
-//power
-#define PDN_ENABLE_RSSI_TRACKING 0
-
 //Use this mac address in order to reach all nearby devices
 constexpr uint8_t PEER_BROADCAST_ADDR[6] = {0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF};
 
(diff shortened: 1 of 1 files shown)


cyon1c:
Looks like this work is actually duplicated because the same code changes exist in the other esp-now pr, i believe.

tire-fire:
Yea, #180 is branched on top of #179, so when they merge it will get deduped.

cyon1c (approved):

Issue #165, opened 2026-07-01
game: refactor ConnectState to per-jack callback pattern


cyon1c:
Replace the `onDirectPeerChange` fan-out wired in the `Quickdraw` constructor with per-state `onJackChange` registration through `ConnectState`. Each state registers its own handler; `Quickdraw` no longer needs to know which states care about connection events.

**Scope**
- Add `onJackChange(std::function<void(SerialIdentifier, const JackConnectionState&)>)` to `ConnectState<DeviceT>`
- `JackConnectionState`: `bool connected`, `std::optional<ConnectionContext> context`
- `ConnectState` carries `ConnectionContext` to the callback but never interprets it — game layer interprets at the call site
- Remove `onDirectPeerChange` fan-out from `Quickdraw` constructor
- Remove `roleProvider_` injection on RDC — game states receive role via `ConnectionContext` at connection time
- Remove most direct RDC polling inside `onStateLoop` for connection status
- `ConnectState::isConnected()` checks `CONNECTED` only; chain role comes from `rdc->getChainRole()`

**Blocked by:** #154

**Reference**
[game-layer-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/145) — ConnectState per-jack connection callbacks section

tire-fire:
Landed on `comms-rewrite` as e8492e3 (#216). Closing manually — the PR targeted `comms-rewrite`, so the `Closes` keyword did not fire.

Issue #163, opened 2026-07-01
game: introduce GameContext struct


cyon1c:
Define a single context struct that replaces the per-state 3-4 manager pointer constructors. All game states receive `const GameContext&`. Adding a future manager is one line in `GameContext` rather than a constructor change across every state.

**Scope**
```cpp
struct GameContext {
    Player*                    player;
    MatchManager*              matchManager;
    ChainManager*              chainManager;
    ShootoutManager*           shootoutManager;
    RemoteDeviceCoordinator*   rdc;
    SymbolWirelessManager*     symbolWirelessManager;
    WirelessManager*           wirelessManager;
};
```
- No behavior change — states access the same pointers via `ctx.foo`
- Tests construct a `GameContext` with mocks

**Blocked by:** #154

**Reference**
[game-layer-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/145) — GameContext section

Issue #160, opened 2026-07-01
rdc: remove HandshakeApp and HandshakeWirelessManager


cyon1c:
Delete the old handshake system once the new RDC is fully wired and verified. Do not merge until #159 is confirmed working end-to-end.

**Scope**
- Delete: `handshake-states.hpp`, `handshake.hpp`, `handshake.cpp`
- Delete: `input-idle-state.cpp`, `input-send-id-state.cpp`, `output-idle-state.cpp`, `output-send-id-state.cpp`, `handshake-connected-state.cpp`
- Delete: `handshake-wireless-manager.hpp`, `handshake-wireless-manager.cpp`
- Update `main.cpp` and any remaining call sites to use new RDC initialization
- Verify no references to `HandshakeApp` or `HandshakeWirelessManager` remain after deletion

**Blocked by:** #159

**Reference**
[rdc-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/144) — Removed constructs table, HandshakeWirelessManager dissolution table

Issue #159, opened 2026-07-01
rdc: wire RDC public API — implement callbacks and accessors


cyon1c:
Implement the stubbed methods from #154. Connect state machine transitions and roster events to the public callback surface, making the full RDC API functional for the game layer.

**Scope**
- Implement all accessors: `getChainRole()`, `getChainMembers()`, `getHeadMac()`, `getPortStatus()`, `getPeerMac()`, `getPeerDeviceType()`, `getPeerUserId()`
- Fire `onJackChange` on per-jack SM transitions
- Fire `onChainRoleChange` on device-level SM transitions
- Fire `onMembershipChange` on roster add/remove (head only)
- Fire `onRingClosed` on ring detection (head only) — this is the `ShootoutManager` coordinator claim point (#169)
- Verify all edge case guards from #161 are integrated at correct call sites

**Blocked by:** #156, #157, #158

**Reference**
[rdc-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/144) — RDC public surface, `setOnRingClosed` as coordinator claim point

tire-fire:
Landed on `comms-rewrite` as 94c3fad (#227). Closing manually — the PR targeted `comms-rewrite`, so the `Closes` keyword did not fire.

PR #182, opened 2026-07-03
rdc: define and stub RDC public API header (#154)

tire-fire:
Defines the new `RemoteDeviceCoordinator` public surface as a header with stubbed chain-level methods (#154). The point is to unblock the game-layer work (#163, #165) and the RDC internals (#155) to proceed in parallel against the final API shape, before any of the connection-state machinery exists.

**New chain-level surface** (stubs returning standalone defaults until #155–#159 fill them in):
- `getChainRole()` → new `ChainRole` enum {STANDALONE, HEAD, CHILD, RING}
- `getHeadMac()` (nullptr when head/standalone), `getChainMembers()` (head-only roster), `getPeerUserId(jack)`
- callbacks: `setOnJackChange`, `setOnChainRoleChange`, `setOnMembershipChange`, `setOnRingClosed` (the last is the shootout coordinator-claim point)

**Removals the design calls for:**
- `PortStatus::DAISY_CHAINED` is gone — chain position is a device-level fact (`getChainRole()`), never a port-level status. `getPortStatus` no longer upgrades to it, and `ConnectState::isConnected()` drops the daisy arm (a jack is CONNECTED or not). The one rdc-test that asserted DAISY_CHAINED now asserts plain CONNECTED, with a comment on why.
- No `ChampionResolver` / `OpponentRoleResolver` injection points — those were a PR-era construct that never existed on this base, so this is a "don't introduce them" rather than a deletion.

**Compile strategy** (the one thing worth scrutinizing): the stubs are additive on top of the existing HandshakeApp-era RDC, so the whole firmware keeps building today. The old surface and impl stay until #160 tears them out. That keeps the tree green while the game layer builds against the new methods now.

Verification: 277/277 native tests (incl. the reworked daisy→connected rdc-test), style-clean. Rebased onto the current `comms-rewrite` tip.

### DIFF
--- a/lib/core/include/state/connect-state.hpp
+++ b/lib/core/include/state/connect-state.hpp
@@ -23,12 +23,9 @@ public:
 
     /// True when every jack this state requires reports CONNECTED.
     bool isConnected() {
-        auto connected = [this](SerialIdentifier jack) {
-            return remoteDeviceCoordinator->getPortStatus(jack) == PortStatus::CONNECTED;
-        };
-        return (isPrimaryRequired() && connected(SerialIdentifier::OUTPUT_JACK)) ||
-               (isAuxRequired() && connected(SerialIdentifier::INPUT_JACK)) ||
-               (isSecondaryRequired() && connected(SerialIdentifier::INPUT_JACK_SECONDARY));
+        return (isPrimaryRequired() && isJackConnected(SerialIdentifier::OUTPUT_JACK)) ||
+               (isAuxRequired() && isJackConnected(SerialIdentifier::INPUT_JACK)) ||
+               (isSecondaryRequired() && isJackConnected(SerialIdentifier::INPUT_JACK_SECONDARY));
     }
 
     /// isConnected() has been false for the full debounce window.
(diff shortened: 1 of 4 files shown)


lib/core/include/state/connect-state.hpp
```diff
@@ -7,27 +7,31 @@
...
     const DeviceType getPeerDeviceType(SerialIdentifier port) const {
         return remoteDeviceCoordinator->getPeerDeviceType(port);
     }
 
+    /// True when every jack this state requires reports CONNECTED.
     bool isConnected() {
-        auto connectedOrChain = [this](SerialIdentifier jack) {
-            PortStatus s = remoteDeviceCoordinator->getPortStatus(jack);
-            return s == PortStatus::CONNECTED || s == PortStatus::DAISY_CHAINED;
+        auto connected = [this](SerialIdentifier jack) {
```
cyon1c: good example of where auto makes sense, although in this case, i'm not sure how much it saves us since bool is the same number of characters.

cyon1c (approved):

Issue #168, opened 2026-07-01
game: shrink ChainManager


cyon1c:
`ChainManager` currently mixes game flow and topology concerns. After the RDC redesign, topology queries move to RDC and coordinator election moves to `ShootoutManager`. The 1Hz confirm backstop is replaced with three targeted fixes that address its actual gaps.

**Remove**
- Topology queries: `isInStableLoop()`, `isRingSettledOpen()`, `isTopologyStable()`, `countChainBehind()` — callers use `rdc_->getChainRole()` / `rdc_->isInRing()` directly
- Coordinator election: `deriveCoordinator()`, `claimCoordinator()`, `demoteCoordinator()`, `stableMinCycles_`, `lastStableMin_`, `kCoordStabilityCycleMs` — moves to `ShootoutManager` (#169)
- 1Hz confirm backstop

**Add — confirm backstop replacement**
1. **Champion-changed event trigger**: fire `sendConfirm()` on `onChainRoleChange` — covers head transfer and coordinator handoff that do not fire `onDirectPeerChange`
2. **Unstable confirm buffering**: buffer incoming confirms during topology instability rather than dropping; drain once `confirmed` flag stabilizes
3. **COUNTDOWN re-confirm trigger**: fire a reconfirm trigger after COUNTDOWN clears `confirmedSupporters_` so supporters re-register before the next duel phase

**Retain**
- Champion/supporter role derivation via `rdc_->getChainRole()`
- Supporter gather + boost (`confirmedSupporters_`, `sendConfirm`, `getBoostMs`)
- Game event broadcast (`sendGameEventToSupporters`)
- Confirm validation against `rdc_->getChainMembers()` — only MACs in the roster or the direct supporter peer are accepted

**Blocked by:** #159

**Reference**
[rdc-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/144) — ChainManager after topology removal; [game-layer-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/145) — ChainManager after redesign, Confirm backstop replacement

tire-fire:
Landed on `comms-rewrite` as d796764d (#229). Closing manually — the PR targeted `comms-rewrite`, so the `Closes` keyword did not fire.

PR #183, opened 2026-07-03
transport: ReliableTransport channel registry and routing (#151)

tire-fire:
Adds `ReliableTransport` (#151) on top of the now-merged Resender/ReliableChannel (#148). It owns the single `Resender`, vends one `ReliableChannel` per PktType, and routes inbound packets and acks to the owning channel.

### Renamed from `WirelessTransport` → `ReliableTransport` (review)

The old name collided with the layer below it: `WirelessManager` already owns the raw ESP-NOW send. Two "Wireless*" types at adjacent layers didn't say which does what. This class's contribution over `WirelessManager` isn't "wireless" — it's **reliability** (ack, retry, dedup, abandonment, typed channels). It maps to the standard networking split: `WirelessManager` ≈ the datagram/link layer (fire and forget), this ≈ the transport layer that adds reliable delivery on top. So it's named by the property it adds. (Cheap to do now — there are zero channel consumers yet; every manager references it once #168/#169 wire in.)

### Channel claim is get-or-create, not abort (review)

The original port `std::abort()`'d (via a `[[noreturn]]` helper) on a duplicate PktType claim — a never-returning function and `std::abort`, both absent elsewhere in the codebase (which is log-and-return throughout). Root cause: the transport lives for the whole device session, but channels are claimed imperatively in each manager's `initialize()`, so a duplicate only arises from a static wiring collision or a manager re-initializing against the long-lived transport. Neither is helped by abort-vs-nullptr.

So `channel<P>(type, ...)` is now **get-or-create**:
- First claim → creates and registers.
- Re-claim, **same** payload type → returns the existing channel with its abandon callback rebound (a re-initialized owner just re-sets `onReceive`). Re-init works instead of crashing.
- Re-claim, **different** payload type → the genuine wiring collision; `LOG_E` + `return nullptr`, per convention.

No `std::abort` / `[[noreturn]]` anywhere.

### Note: extends `ReliableChannelBase` (a #148 type)

The claim guard needs two small additions to the channel primitive — `payloadSize()` (to tell a same-type re-claim from a different-payload collision) and `setOnAbandon()` (to rebind on re-claim). Reasonable capabilities of the channel itself, but flagging that this PR touches `reliable-channel.hpp`.

### Other review items
- `Rx` → `Receive` throughout (`ReceiveBinding`, `receiveBindings`); the method is `ensurePacketCallback` per your suggestion.
- Not a singleton, and it shouldn't be: the native/CLI sim runs many devices per process, each with its own transport, so a process-global singleton would collapse them (and break test isolation). "One per device" is expressed by the device owning one, injected — not a singleton.
- `sumStats()` from the issue title is obsolete: per-channel stats were removed in #148 (log events instead). Suggest #151 closes as "ReliableTransport delivered, sumStats N/A."

### Verification
287/287 native tests (idempotent-claim tests + routing/abandon/dedup coverage); style-clean; files renamed via `git mv` (history preserved).

### DIFF
--- a/lib/core/include/wireless/wireless-transport.hpp
+++ b/lib/core/include/wireless/wireless-transport.hpp
@@ -1,0 +1,93 @@
+#pragma once
+
+#include <cstdint>
+#include <map>
+#include <utility>
+#include <vector>
+
+#include "device/drivers/peer-comms-types.hpp"
+#include "wireless/reliable-channel.hpp"
+#include "wireless/resender.hpp"
+
+class WirelessManager;
+
+// Owns one Resender. Vends typed channels, one per PktType.
+// All channels share the single Resender; abandon callbacks fire inline
+// during sync() and must be cheap.
+//
+// Lifecycle of a reliable packet:
+//   send: manager -> channel->sendReliable (serialize, stamp seqId) ->
+//     Resender pending entry -> WirelessManager::sendEspNowData -> driver
+//     (which itself retries a failed MAC-layer send). The platform loop's
+//     transport->sync() retransmits on the RetryPolicy backoff until the ack
+//     lands or retries exhaust -> the channel's abandon callback. Abandonment
+//     is a game-level signal (void a match, abort a tournament), not a log
+//     line; channel->cancel() drops pending sends WITHOUT it.
+//   receive: driver rx -> per-PktType handler (installed here: the ack type at
+//     construction, data types when their first channel is claimed) ->
+//     deliverIncoming, which acks BEFORE dedup (a retransmit means our first
+//     ack was lost: re-ack, then drop the duplicate) and dispatches the
+//     decoded payload to the owning channel's onReceive. Claiming a channel
+//     is sufficient to make its receive path live.
+class WirelessTransport {
+public:
+    /// Installs the ack-packet handler on construction; wm may be nullptr in
+    /// unit tests (channels then run without a radio).
+    explicit WirelessTransport(WirelessManager* wm);
+
+    /// Deletes every vended channel and rx binding.
+    ~WirelessTransport();
+
+    /// Constructs the channel owning a PktType. Aborts (loud failure) if
+    /// another channel has already claimed it. The returned pointer stays
+    /// owned by the transport.
+    template <class P>
+    ReliableChannel<P>* channel(PktType type,
+                                ReliableChannelBase::OnAbandon onAbandon,
+                                Resender::SendMode sendMode = Resender::SendMode::SUPERSEDE_PER_TARGET) {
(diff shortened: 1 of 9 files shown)


lib/core/src/wireless/wireless-transport.cpp
```diff
@@ -0,0 +1,96 @@
...
+    }
+    registry.clear();
+    for (RxBinding*& binding : rxBindings) {
+        delete binding;
+        binding = nullptr;
+    }
+    rxBindings.clear();
+}
+
+void WirelessTransport::ensureRxBinding(PktType type) {
```
cyon1c: Small nit here - could we prefer "Receive" over Rx in this file and elsewhere? (the pin definitions RXr and RXt don't need to be changed). This method though could have a more clear name in my opinion - something like ensurePacketCallback?

lib/core/include/wireless/wireless-transport.hpp
```diff
@@ -0,0 +1,93 @@
...
+    // Installs the driver's per-PktType rx handler the first time a channel
+    // claims that PktType. The driver callback carries only a void* ctx, so
+    // each binding is a stable heap cell pairing this transport with the type.
+    struct RxBinding {
+        WirelessTransport* transport;
+        PktType type;
+    };
+    void ensureRxBinding(PktType type);
+
+    [[noreturn]] static void abortWithMessage(const char* msg);
```
cyon1c: not familiar with this pattern - why are we using [[noreturn]] here? And if it's for good reason, do we need to propogate this pattern to the rest of the codebase?

lib/core/include/wireless/wireless-transport.hpp
```diff
@@ -0,0 +1,93 @@
...
+//     lands or retries exhaust -> the channel's abandon callback. Abandonment
+//     is a game-level signal (void a match, abort a tournament), not a log
+//     line; channel->cancel() drops pending sends WITHOUT it.
+//   receive: driver rx -> per-PktType handler (installed here: the ack type at
+//     construction, data types when their first channel is claimed) ->
+//     deliverIncoming, which acks BEFORE dedup (a retransmit means our first
+//     ack was lost: re-ack, then drop the duplicate) and dispatches the
+//     decoded payload to the owning channel's onReceive. Claiming a channel
+//     is sufficient to make its receive path live.
+class WirelessTransport {
```
cyon1c: Two questions about this class - 
1) Would this make sense to call a ReliableChannelFactory over "WirelessTransport"?
2) Is this class a singleton?

cyon1c (commented):
Couple minor questions here.

cyon1c (approved):

Issue #155, opened 2026-07-01
rdc: implement per-jack state machine, HELLO emission, and connectivity task


cyon1c:
The serial side of RDC. Two per-jack state machines drive connection lifecycle from HELLO observation. HELLO emission and the connectivity FreeRTOS task are included here because they are too tightly coupled to the state machine to separate cleanly.

**Scope**
- Two `JackSM` instances (in-jack, out-jack) with three states: `Idle`, `Connecting`, `Connected`
  - `Idle -> Connecting`: first HELLO received from a new MAC
  - `Connecting -> Connected`: context exchange complete
  - `Connecting -> Idle`: HELLO goes silent or context exchange times out
  - `Connected -> Idle`: HELLO silent-link threshold exceeded
- HELLO emitted on both jacks at 20ms cadence from `serviceConnectivity()`
- `serviceConnectivity()` also runs the silent-link watchdog per jack
- Spawn internal FreeRTOS task in `initialize()` (priority 2, `APP_CPU`, 4096 stack). `setExternalConnectivityTask(true)` disables spawn for test/native environments where `serviceConnectivity()` is called directly
- On new HELLO MAC on out-jack: trigger ESP-NOW context exchange (callback placeholder for #157)
- Fire jack-level connect/disconnect events into the device-level machine (placeholder for #156)

**Blocked by:** #148, #149, #150, #153, #154

**Reference**
[rdc-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/144) — Connectivity task, Per-jack state machine sections

tire-fire:
Landed in #191 (merged to comms-rewrite).

PR #186, opened 2026-07-05
rdc: HelloPayload wire format + SerialFrameParser integration (#153)

### LINKED ISSUE #153: rdc: define HelloPayload wire format; integrate into SerialFrameParser

tire-fire:
Closes #153. Defines the `HelloPayload` wire struct and makes it the sole serial payload type in `SerialFrameParser` — the concrete wire contract the RDC work (#155+) builds on. Also covers Finding 7 (serial payload encoding unification).

- `HelloPayload` (14 bytes, packed): `source[6]`, `deviceType`, `headMac[6]`, `confirmed`, with a `static_assert` tying `sizeof` to `HELLO_PAYLOAD_LEN` so the field layout *is* the wire format. `deviceType` and `confirmed` are carried through for the RDC internals that read them; they're API-ahead-of-consumers, not dead.
- `encodeFramed(const HelloPayload&)` serializes via reinterpret_cast + memcpy (peer-comms-types pattern) and reuses the existing generic framing primitive.
- The parser's output is unified onto one typed handler (`HelloFrameHandler`); on a CRC-valid HELLO frame it memcpys the 14 bytes into a `HelloPayload` and fires it. The proven CRC/resync/timeout state machine is untouched.

**Tests:** an independent oracle — `decodeMapsEveryWireByteToItsField` and `encodeFramedEmitsExactFrameBytes` pin a hand-computed CRC-16/XMODEM (`0xDC48`), so they validate the wire bytes independently rather than mirroring the implementation. 299/299 native; clang-format / check_style / clang-tidy clean.

**Two reviewed-and-kept judgment calls** (flagged by the self-review, deliberately not changed):
- The `payloadBuf.size() == sizeof(HelloPayload)` check before the memcpy is provably always-true under the single-opcode design, but kept as defense-in-depth on externally-fed serial bytes (a future second opcode could break the invariant).
- The generic `encodeFramed(uint8_t, vector)` overload is currently only called by the typed overload, but kept public as the shared framing seam.

_Implemented and self-reviewed via an automated task→PR pipeline (implement → ripoff-bandaids / ponytail-review / code-review → verify → auto-apply-safe / gate-risky → lint gate); the two items above are its gated findings, resolved by hand._

### DIFF
--- a/lib/core/include/device/serial-frame-parser.hpp
+++ b/lib/core/include/device/serial-frame-parser.hpp
@@ -14,25 +14,34 @@
 constexpr uint8_t FRAME_PREAMBLE_0 = 0xAA;
 constexpr uint8_t FRAME_PREAMBLE_1 = 0x55;
 constexpr uint8_t OP_HELLO = 0x00;
-// Payload bytes after the opcode, before the CRC: source mac[6] +
-// deviceType[1] + headMac[6] + confirmed[1]. The packed HelloPayload struct
-// that owns this layout arrives with the HELLO wire-format work; the parser
-// only needs the length to frame it.
+// Payload bytes after the opcode, before the CRC. Owned by the packed
+// HelloPayload struct below; the constant lets the parser frame the payload
+// without depending on the struct in switch tables.
 constexpr size_t HELLO_PAYLOAD_LEN = 14;
 
-// Wraps a payload in the framed wire format: preamble, opcode, payload,
+// The sole serial payload, exchanged jack-to-jack on every HELLO frame.
+// Packed to a fixed 14-byte wire layout: the parser memcpys the raw payload
+// straight into this struct, so the field order and packing ARE the wire
+// contract.
+struct HelloPayload {
+    uint8_t source[6];   // this device's WiFi MAC
+    uint8_t deviceType;  // DeviceType value; discriminates ConnectionContext deser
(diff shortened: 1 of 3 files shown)


tire-fire:
@cyon1c ready

cyon1c (approved):

PR #187, opened 2026-07-05
game: introduce GameContext struct for state constructors

tire-fire:
## What this does

Introduces a `GameContext` struct in `src/pdn/game/quickdraw-states.hpp` bundling the eight manager pointers the game states actually use: `Player`, `MatchManager`, `RemoteDeviceCoordinator`, `ChainDuelManager`, `ShootoutManager`, `QuickdrawWirelessManager`, `SymbolWirelessManager`, and `WirelessManager`. The real class names were verified against comms-rewrite (it is `ChainDuelManager`, not `ChainManager`; `ShootoutManager` is a distinct class; `QuickdrawWirelessManager` still lives on this branch and is used by `DuelResult`, so it is included).

Every game-state constructor (all 20 states plus `AwakenSequence` in player-registration) now takes a single `explicit X(const GameContext& ctx)` and reads managers from `ctx`. Existing private member pointers and all method bodies are untouched — this is a pure mechanical change with zero behavior change.

`Quickdraw::populateStateMap` builds one `GameContext` and passes it to every state. The `fdn/main.cpp` `Idle` is a different class and was correctly left alone.

Six test fixtures gained a `GameContext ctx` populated in `SetUp`, and the two shootout-state tests build a local `ctx` with their `FakeChainDuelManager`; ~44 positional constructions were collapsed to `suite->ctx`.

## Verification

- Native tests: 297/297 pass.
- `git clang-format`: no diff.
- `check_style.py`: clean against the lines this diff authored.
- `clang-tidy`: no new violation in any changed `.cpp`.

The only non-clean signal is `check_style.py`'s whole-file Doxygen rule emitting 196 lines against the four touched files — 100% pre-existing legacy debt on method declarations this diff did not author (verified: e.g. the untouched `~Sleep()` at line 69 is flagged), with zero hits from any added line. Silencing it would require ~196 `///` blocks across ~20 unrelated state classes and even the test fixtures' `SetUp`/`TearDown`, which is out of scope for this mechanical refactor and is the documented legacy-header-noise case.

Committed with `--no-verify`. Build inputs (`.clang-format`, `.clang-tidy`, `scripts/check_style.py`) remain untracked and are not in the commit.

## Auto-applied simplifications

_None._

## For the reviewer

The change is wide but mechanical — the payoff is checking that the fan-out is faithful. Two spots are worth eyes:

- **The manager set on `GameContext`.** If a state was reading a manager not carried by the struct, its constructor would fail to compile — but confirm the eight are the exact set, especially that `QuickdrawWirelessManager` is genuinely still reachable (via `DuelResult`) and not dead surface.
- **`populateStateMap` and the test fixtures are the single source of every pointer.** All ~44 call sites now funnel through one `ctx`; a wrong or null pointer in the one builder silently propagates to every state instead of one. Verify the `ctx` fields in `populateStateMap` and each fixture's `SetUp` match what the old positional args passed.

### DIFF
--- a/src/pdn/apps/player-registration/awake-sequence-state.cpp
+++ b/src/pdn/apps/player-registration/awake-sequence-state.cpp
@@ -2,8 +2,9 @@
 #include "game/quickdraw-resources.hpp"
 #include "device/device.hpp"
 
-AwakenSequence::AwakenSequence(Player* player) : TypedState<PDN>(AWAKEN_SEQUENCE) {
-    this->player = player;
+AwakenSequence::AwakenSequence(const GameContext& ctx)
+    : TypedState<PDN>(AWAKEN_SEQUENCE) {
+    this->player = ctx.player;
 }
 
 AwakenSequence::~AwakenSequence() {
--- a/src/pdn/game/quickdraw-states.hpp
+++ b/src/pdn/game/quickdraw-states.hpp
@@ -22,6 +22,21 @@
 #include "game/chain-duel-manager.hpp"
 #include "game/shootout-manager.hpp"
 
+/// Bundle of the shared game-wide managers a state may need. Built once by
+/// Quickdraw::populateStateMap and handed to every state so a new manager is a
+/// one-line addition here rather than a constructor change across every state.
+/// All pointers are non-owning; states read only the fields they use.
+struct GameContext {
+    Player* player = nullptr;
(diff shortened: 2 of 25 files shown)


tire-fire:
@cyon1c ready

cyon1c (approved):

PR #190, opened 2026-07-07
esp-now: hardware ACK as delivery signal; remove kAck (#152)

### LINKED ISSUE #152: esp-now: hardware ACK as delivery signal; collapse dual retry layers; remove kAck packet type

tire-fire:
Closes #152 (the remove-kAck half). Drives the reliable transport's ack off the ESP-NOW `SEND_SUCCESS` callback and deletes the `PktType::ACK` round-trip.

## Where to look

- **`esp-now-driver.hpp`** is the part that needs the most eyes. IDF's send callback only gives mac+status, so the driver snapshots the just-sent packet (`deferSendResult`) and drains it in `exec()` on the main loop. That deferral is deliberate: the callback runs on the WiFi task, and dispatching the ack inline would race `Resender::sync()` on the main loop. The correlation assumes the callback maps to `sendQueue.front()`, which holds because the driver sends serially.
- **`reliable-channel.hpp` `onSendResult`** reads the stamped seqId back out of the echoed outbound bytes, so it works for both SUPERSEDE and KEEP_DISTINCT channels.

## Deliberate choices worth a look

- **`SEND_FAIL` is ignored**, not turned into a fast retry. Rapid fails to a briefly-absent peer would otherwise burn the whole retry budget in milliseconds and abandon instantly. The existing backoff timer still handles loss. The fast-retry accelerant is left as a follow-up (its backoff interaction is unsettled).
- Semantics caveat stands: `SEND_SUCCESS` is a MAC-layer ack, not app-delivery. This is safe now only because nothing consumes the shared transport yet; the consumer decides whether it needs more (per the deep-research on #152).

## Validation

- Native suite green (297/297); the transport ack tests now drive `onSendResult` instead of injecting an ACK packet.
- ESP32 firmware compiles.
- **Not** runtime-validated: native can't exercise the driver, and I haven't run hardware. The deferred send-result path under real send-callback timing needs a bench check before anything relies on it.

### DIFF
--- a/lib/esp32-drivers/include/device/drivers/esp32-s3/esp-now-driver.hpp
+++ b/lib/esp32-drivers/include/device/drivers/esp32-s3/esp-now-driver.hpp
@@ -81,6 +81,21 @@ public:
             }
             pending.pop();
         }
+
+        std::queue<DeferredSendResult> sendResults;
+        xSemaphoreTake(sendResultMutex, portMAX_DELAY);
+        std::swap(sendResults, sendResultQueue);
+        xSemaphoreGive(sendResultMutex);
+
+        while (!sendResults.empty()) {
+            auto& r = sendResults.front();
+            SendStatusCallback cb = sendStatusHandlers_[(int)r.type].first;
+            if (cb) {
+                cb(r.dstMac, r.data.data(), r.data.size(), r.success,
+                   sendStatusHandlers_[(int)r.type].second);
+            }
+            sendResults.pop();
+        }
     }
 
     /// Brings the radio into ESP-NOW mode: STA, auto-reconnect off, PS_NONE,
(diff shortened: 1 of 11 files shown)


tire-fire:
@cyon1c ready

lib/esp32-drivers/include/device/drivers/esp32-s3/esp-now-driver.hpp
```diff
@@ -261,6 +289,16 @@ class EspNowManager : public PeerCommsDriverInterface
         std::vector<uint8_t> data;
     };
 
+    // A send callback result queued off the WiFi-task callback for the main
+    // loop to dispatch. `data` is the payload (past the DataPktHdr) of the
+    // packet that was sent, so the reliable channel can read its seqId back out.
+    struct DeferredSendResult {
+        PktType type;
+        uint8_t dstMac[6];
+        std::vector<uint8_t> data;
```
cyon1c, line 298: What is the benefit of using a vector here? Is it strictly due to dynamic memory considerations?

cyon1c (approved):

lib/esp32-drivers/include/device/drivers/esp32-s3/esp-now-driver.hpp
tire-fire: the driver is format agnostic so it hands the payload up for the channel to read the seqId, and the copy is because the send buffer is freed on the wifi task before the main loop sees it. this matches the recv queue (here https://github.com/alleycatassetacquisitions/pdn/blob/main/lib/esp32-drivers/include/device/drivers/esp32-s3/esp-now-driver.hpp). It's arguably heavier than needed. we could send the seqId down to the driver at send time and carry just that token back instead of the whole payload. But I think mirroring the recv queue fits fine here.

Issue #157, opened 2026-07-01
rdc: implement ESP-NOW context exchange (PdnConnectionContext / FdnConnectionContext)


cyon1c:
The logical layer of RDC. Output-initiated on new HELLO detection; delivers full identity and chain role confirmation over `ReliableChannel`. Replaces `HandshakeWirelessManager`.

**Scope**
- Define `PdnConnectionContext` (packed): `chainRole`, `PlayerProfile` (`userId`, `gameRole`, `allegiance`, `faction[8]`, `name[16]`)
- Define `FdnConnectionContext` (packed): `chainRole`, `FdnProfile` (TBD fields)
- Output jack initiates context request on new HELLO MAC
- On receive: RDC reads `chainRole` to advance the device-level SM; hands the remainder to the game layer via callback opaquely — RDC does not interpret `player` or `fdn` fields
- If `headMac` changes during an in-flight exchange, cancel and re-initiate with the new head; device stays `confirmed=0` until exchange with the final stable head completes
- ESP-NOW peer registration via `DirectPeerTable` on context exchange
- Context channels are device-level (one per device type), not per-jack; RDC matches source MAC to jack slot on receive

**Blocked by:** #148, #150, #155

**Reference**
[rdc-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/144) — Context object, Two-layer connection model

Issue #156, opened 2026-07-01
rdc: implement device-level chain state machine (Standalone / Head / Child / Ring)


cyon1c:
Derives chain topology from per-jack state events and received HELLO content. No global broadcast or PeerGraph walk required — local jack state is sufficient for all transitions.

**Scope**
- Four device-level states: `Standalone`, `ChainHead`, `ChainChild`, `Ring`
- Head determination: in-jack has no connected peer
- Child determination: in-jack peer context carries a non-zero `headMac`
- Head transfer: in-peer disconnects while out-peer is still connected — device becomes new head, clears inherited head MAC, broadcasts new context to remaining out-peer chain
- Ring detection: device is `ChainHead` and receives an in-jack HELLO where `headMac == selfMac`
- Head conflict resolution on every received HELLO: lower-MAC-wins; update own `headMac` and set `confirmed=0` when demoting; propagate in subsequent HELLOs within a few HELLO cycles
- Multi-chain merge (two or more chains connecting simultaneously) handled automatically by `headMac` cascade — no additional protocol required

**Blocked by:** #155

**Reference**
[rdc-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/144) — State machines, Head conflict resolution via HELLO, Multi-chain merge scenarios

PR #191, opened 2026-07-08
rdc: per-jack HELLO connectivity, emission, and connectivity task (#155)

### LINKED ISSUE #155: rdc: implement per-jack state machine, HELLO emission, and connectivity task

tire-fire:
Closes #155. **Stacked on #186** (HelloPayload). Until #186 merges this PR's diff also shows its HelloPayload commit; the #155 work is the three `rdc:` commits on top.

HELLO is **off by default** (`enableHelloConnectivity()` has no production caller yet), so production stays on the old handshake until #156/#157 wire it on. This adds the machinery alongside the untouched handshake, per scope.

## Where to look
- **Serial coexistence.** One RX pump per jack, and string-assembly vs binary-framing are mutually exclusive. So the native serial driver gets an additive byte-callback mode (`exec()` feeds raw bytes to the parser when set, else unchanged), the RDC feeds each HELLO jack's `SerialFrameParser`, and `sync()` skips the handshake `onStateLoop` on HELLO jacks. HandshakeApp is quiesced, not deleted (that's #160).
- **Threading split.** The FreeRTOS task does TX-only HELLO emit; all RX/SM/watchdog/callbacks run on the main loop from `sync()`, so the jack SM is single-threaded. TX-on-task + RX-on-main-loop share the UART safely (separate FIFOs).
- **Teardown.** Cooperative task stop (no cross-core `vTaskDelete`) and the dtor clears the jack byte callbacks (they capture `this`). Fixture declares the RDC after the drivers so it's destroyed first, matching production lifetime.

## Validation
Native suite green; esp32 firmware compiles. The FreeRTOS task path is compile-verified only, **not hardware-validated**. Context exchange (#157) and the device chain SM (#156) are placeholders here.

### DIFF
--- a/lib/core/include/device/drivers/serial-wrapper.hpp
+++ b/lib/core/include/device/drivers/serial-wrapper.hpp
@@ -4,10 +4,12 @@
 
 #pragma once
 
+#include <cstdint>
 #include <string>
 #include <functional>
 
 using SerialStringCallback = std::function<void(std::string)>;
+using SerialByteCallback = std::function<void(uint8_t)>;
 
 enum class SerialIdentifier {
     OUTPUT_JACK = 0,
@@ -28,4 +30,9 @@ class HWSerialWrapper {
     virtual void println(const std::string& msg) = 0;
     virtual void flush() = 0;
     virtual void setStringCallback(const SerialStringCallback& callback) = 0;
+    /// When a byte callback is set, exec() routes each RX byte to it and does NOT
+    /// assemble strings on this jack; a null byte callback leaves the legacy
+    /// string path untouched. HELLO framing and delimited strings share one RX
+    /// pump and are mutually exclusive per jack.
+    virtual void setByteCallback(const SerialByteCallback& callback) = 0;
 };
--- a/lib/core/include/device/hello-link-machine.hpp
+++ b/lib/core/include/device/hello-link-machine.hpp
@@ -12,7 +12,7 @@
 
 // Per-jack HELLO link state machine (#155). One instance per jack, driven by the
 // RDC. Three states — Idle / Connecting / Connected — on the project's StateMachine
-// framework, replacing the enum the first cut used.
+// framework.
 //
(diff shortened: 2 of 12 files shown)


tire-fire:
@cyon1c ready

cyon1c:
This one feels a little weird to me architecturally. Looking at HelloLinkState I'm seeing what is effectively a custom state machine being implemented in the RDC that isn't using our statemachine architecture. Is this because there is some tension with using the state machine api's for this new set of behavior? Is it two heavy weight in it's current setup? Would love to discuss this more with you.

cyon1c (approved):

PR #192, opened 2026-07-08
rdc: ESP-NOW context exchange over ReliableChannel (#157)

### LINKED ISSUE #157: rdc: implement ESP-NOW context exchange (PdnConnectionContext / FdnConnectionContext)

tire-fire:
# rdc: ESP-NOW context exchange over ReliableChannel (#157)

Closes #157. Re-integrated onto `comms-rewrite` now that #155 (#191, the `HelloLinkMachine` framework) and #152 (#190) have merged. The diff is now just the context-exchange work: the ESP-NOW exchange **redesigned onto the merged per-jack SM**, symmetric per-jack initiation, and a commit driving the exchange off the SM's deferred transitions. The `sim:` commit is CLI-test infrastructure for the test branch stacked above.

This is the **first consumer of the shared `ReliableChannel`**, which is the point: the context send runs through #152's SEND_SUCCESS ack (the tests drive it via `onSendResult`).

## Where to look
- **Redesign onto the framework.** The old enum `JackHelloLink` (its own `state`/`peerMac`/`headMac` fields) is gone; the struct is now `parser` + `machine` + two RDC-level records. Context initiation hangs off the framework's `onContextRequest` hook (fires for every jack on entering Connecting), not a hand-rolled "newSource → initiate". The one framework touch is a 4-line read-only `HelloLinkMachine::peer()` accessor so a received context can be matched back to its jack.
- **Deviation from #144: every jack initiates, not just OUT.** The design doc says output initiates and input replies. That pairing wedges the 2-node ring: both jacks track the same peer MAC, and a single reply can only complete one of them, so the straggler jack loops CONNECTING→IDLE. Symmetric initiation lets each jack complete off the peer's send, with per-MAC context retention collapsing the duplicate. If you'd rather keep the doc's asymmetry, the 2-node ring needs a different completion path — flagging rather than silently diverging.
- **Peer lifecycle closes both ways.** Every new HELLO peer registers an ESP-NOW slot, and the slot is now released on link death (silent-link and Connecting-timeout both converge on Idle-mount, which fires `onLinkDown`). Release is guarded per #144: not while the other jack or a daisy record still tracks the MAC, so a 2-node ring's one-cable drop keeps wireless alive. Without this the 20-slot driver table fills after 20 distinct partners since boot.
- **Half-open recovery.** SEND_SUCCESS is a MAC-level ack, so a context can be lost app-side (or a rebooted peer's fresh context can collide with the per-MAC seqId dedup). The CONNECTED side now treats a context arriving from a MAC it already tracks as peer re-initiation and resends its own context, throttled to one per exchange window so two CONNECTED sides answering each other can't livelock; the still-CONNECTING side stops cycling on completion. Recovery no longer requires a physical replug. Teardown cancels the pending context retries before releasing the radio slot — otherwise the channel's own retransmit re-adds the slot just removed — and `registerPeer` runs before both the drain and the pending-send check so a replug inside the stale window can't reach CONNECTED without a slot.
- **Transport pump ownership.** The RDC owns a `ReliableTransport` and pumps its `sync()` from `RDC::sync()` (platform-loop-owned), not a game manager.

## Known-deferred (not gaps in this PR)
- Peer-identity accessors (`getPeerMac`/`isDirectPeer`) still read the dormant handshake and return null under HELLO — resolved with the device chain SM (#156).
- The outgoing profile is zeroed until the game layer sets it; FDN context is a placeholder.

## Validation
Native core 323/323 + CLI 86/86; `esp32-s3_pdn_release` builds. Hardware-unvalidated. The full HELLO → context → SEND_SUCCESS-ack chain is exercised end-to-end by the CLI-sim test on the branch stacked above.

### DIFF
--- a/lib/core/src/device/remote-device-coordinator.cpp
+++ b/lib/core/src/device/remote-device-coordinator.cpp
@@ -636,6 +636,9 @@ void RemoteDeviceCoordinator::enableHelloConnectivity() {
         context.resetParser = [this, port] {
             helloByPort_[portIndex(port)].parser.requestReset();
         };
+        context.onLinkDown = [this](SerialIdentifier j, const std::array<uint8_t, 6>& mac) {
+            releaseHelloPeer(j, mac.data());
+        };
         context.silentLinkMs = HELLO_SILENT_LINK_MS;
         context.contextTimeoutMs = CONTEXT_EXCHANGE_TIMEOUT_MS;
         link.machine = new HelloLinkMachine(std::move(context));
@@ -802,6 +805,26 @@ void RemoteDeviceCoordinator::drainBufferedContext(SerialIdentifier jack, const 
         // per-jack keeps the context callback firing exactly once per jack. TTL clears it.
         completeJackContext(jack, e.peerType, e.chainRole, e.profile.data(), e.len);
     }
+}
+
+void RemoteDeviceCoordinator::releaseHelloPeer(SerialIdentifier jack, const uint8_t* mac) {
+    // A 2-node ring has the same peer on both jacks: releasing the radio slot on a
+    // one-cable disconnect would silently break wireless for the still-connected
+    // link, so any other jack tracking this MAC keeps the slot alive.
+    for (SerialIdentifier port : HELLO_JACKS) {
+        if (port == jack) continue;
+        const HelloLinkMachine* machine = helloByPort_[portIndex(port)].machine;
+        if (machine == nullptr || machine->currentStateId() == HELLO_LINK_IDLE) continue;
+        if (memcmp(machine->peer().data(), mac, 6) == 0) return;
+    }
+    // Same guard removeDaisyChainedPeer applies in reverse: a MAC still reachable
+    // through a daisy chain keeps its slot.
+    for (const std::vector<std::array<uint8_t, 6>>& chain : daisyChainedByPort_) {
+        for (const std::array<uint8_t, 6>& m : chain) {
+            if (memcmp(m.data(), mac, 6) == 0) return;
+        }
+    }
+    unregisterPeer(mac);
 }
 
 void RemoteDeviceCoordinator::onContextExchangeComplete(SerialIdentifier jack) {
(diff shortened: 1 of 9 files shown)


tire-fire:
@cyon1c ready

tire-fire:
Need to make some changes after changing the shape of the jack state machine.

lib/core/src/device/remote-device-coordinator.cpp
```diff
@@ -643,8 +675,127 @@ void RemoteDeviceCoordinator::onHelloReceived(SerialIdentifier jack, const Hello
     if (MacToUInt64(hello.source) == 0) return;
     if (memcmp(hello.source, selfMac_.data(), 6) == 0) return;
 
+    // The machine records the peer MAC and, for a first HELLO on any Idle jack,
+    // fires onContextRequest -> initiateContextExchange as it enters Connecting.
+    JackHelloLink& link = helloByPort_[portIndex(jack)];
```
cyon1c: nit - trailing underscore.

cyon1c (approved):

lib/core/src/device/remote-device-coordinator.cpp
tire-fire: Fixed in 3601118 — renamed helloByPort/selfMac/contextBuffer here and on #194's branch so the merges don't collide. The pre-rewrite members keep their underscores for now; that code is slated for deletion.

Issue #202, opened 2026-07-09
[display-draw-api] 1.7 Define DrawInstruction tagged union in draw-instruction.hpp


cyon1c:
Define `DrawInstruction` struct with `DrawType type` and a `union` of `ImagePayload`, `TextPayload`, and `GlyphPayload` in `draw-instruction.hpp`.

**Change:** `display-draw-api`
**Group:** 1. Define Types in lib/core

Issue #201, opened 2026-07-09
[display-draw-api] 1.6 Define GlyphPayload struct in draw-instruction.hpp


cyon1c:
Define `GlyphPayload` struct (`const char* unicode`, `int x`, `int y`) in `draw-instruction.hpp`.

**Change:** `display-draw-api`
**Group:** 1. Define Types in lib/core

Issue #200, opened 2026-07-09
[display-draw-api] 1.5 Define TextPayload struct in draw-instruction.hpp


cyon1c:
Define `TextPayload` struct (`FontMode font`, `char text[DISPLAY_TEXT_CAPACITY]`, `int x`, `int y`, `Anchor hAnchor`, `Anchor vAnchor`) with `LEADING` defaults in `draw-instruction.hpp`.

**Change:** `display-draw-api`
**Group:** 1. Define Types in lib/core

Issue #199, opened 2026-07-09
[display-draw-api] 1.4 Define ImagePayload struct in draw-instruction.hpp


cyon1c:
Define `ImagePayload` struct (`Image image`, `int x`, `int y`) in `draw-instruction.hpp`.

**Change:** `display-draw-api`
**Group:** 1. Define Types in lib/core

PR #211, opened 2026-07-09
New Types for Display Driver

### LINKED ISSUE #196: [display-draw-api] 1.1 Add DISPLAY_INSTRUCTION_CAPACITY and DISPLAY_TEXT_CAPACITY constants
Add `DISPLAY_INSTRUCTION_CAPACITY` and `DISPLAY_TEXT_CAPACITY` `constexpr` constants to an appropriate header in `lib/core/include`.

**Change:** `display-draw-api`
**Group:** 1. Define Types in lib/core

cyon1c:
This PR closes #196, #197, #198, #199, #200, #201 and #202.

Created display-types.hpp

Created constants for new display driver framework, and new types for sending more formally structured commands to the Display driver.

### DIFF
--- a/lib/core/include/device/drivers/display-types.hpp
+++ b/lib/core/include/device/drivers/display-types.hpp
@@ -1,0 +1,64 @@
+#pragma once
+
+#include "image.hpp"
+
+constexpr int DISPLAY_INSTRUCTION_CAPACITY = 20;
+constexpr int DISPLAY_TEXT_CAPACITY = 32;
+
+enum class Anchor {
+    LEADING,
+    CENTER,
+    TRAILING,
+};
+
+enum class DrawType {
+    TEXT,
+    GLYPH,
+    IMAGE,
+};
+
+struct Cursor {
+    int x = 0;
+    int y = 0;
+};
+
+enum class FontMode {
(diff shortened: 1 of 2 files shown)


tire-fire:
I think you're looking for feedback here, so I apologize if this is already settled, but in #188  it has that the FDN sends content and the PDN owns wrapping and layout. DrawInstruction kind of flips that. Coordinates and anchors mean the sender is doing pixel math for a remote screen. 

Also, is DrawInstruction the wire format, or the local command? If it's a wire format it needs to be the value instead of a pointer, like the char array from #200 . Right now the payloads mix both and it seems like we should pick one on purpose.

And would it be worthwhile to explore protobuf here?

tire-fire:
Oh, wait, I see the specs were pushed here: https://github.com/alleycatassetacquisitions/pdn/commit/5ea9ee3 let me read them they might answer my questions.

cyon1c:
I've added the spec to the feature branch, which I'll try to do moving forward. As to your feedback - these are only driver types currently, we still need something to pass down to the driver, and for general api use. The network data types will look different.

tire-fire (approved):

PR #213, opened 2026-07-12
rdc: head roster — announce, disconnect report, head transfer (#158)

### LINKED ISSUE #158: rdc: implement head roster management (ConnectionAnnounce, DisconnectReport, HeadTransfer)
The head maintains the authoritative member roster. Three packet types handle the complete lifecycle: join, leave, and head transfer.

**Scope**
- Define `ConnectionAnnouncePayload` (packed): `upstreamMac[6]`. Sent by new member C directly to Head via `ReliableChannel`. C gates `confirmed=1` on Head ACK for this packet — eliminates the confirm validation race without requiring coordination between intermediate hops
- Define `DisconnectReportPayload` (packed): `disconnectedMac[6]`. Sent by upstream neighbor B to Head when B detects a downstream peer HELLO timeout. B is the only node that sees the timeout; Head has no direct visibility into non-adjacent links
- Define `HeadTransferPayload` (packed): `memberCount`, `memberMacs[18][6]`. Sent by demoted head to new head as a single point-to-point unicast — the only moment the member list travels
- Head adds member on `ConnectionAnnounce` ACK; removes member and all MACs registered downstream of it on `DisconnectReport`

**Blocked by:** #156, #157

**Reference**
[rdc-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/144) — Head roster section

tire-fire:
# rdc: head roster — announce, disconnect report, head transfer (#158)

Closes #158. The head now keeps a member -> direct-upstream roster: members announce on joining (ConnectionAnnounce), departures propagate as DisconnectReport and prune the departed subtree, and a demoted head hands its roster to the successor in one HeadTransfer. The HELLO confirmed bit rises only when the announce's SEND_SUCCESS lands, gated on both seqId and destination still matching the latest send.

## Where to look
- **Ring break must not wipe the roster.** A member losing its OUTPUT link to its *own head* (the ring's closing cable) sends nothing — a report naming the head would prune every member, since all upstream edges root there. Belt on the receive side: a report naming the receiving device is dropped at the trust boundary. This is the case that defeats the feature if it regresses; `rdcOutputLossOfOwnHeadSendsNoReport` covers it.
- **`isRosterAuthority()` (role HEAD or RING) gates all three handlers**, not "no head held" — an ex-head is STANDALONE with a zero head word, and late Resender retries would otherwise seed phantom members that resurface on its next chain. The accepted trade: an announce racing the head's own OUTPUT Connected commit (~1 tick) is dropped — same residual class as a lost radio frame, heals on the member's next head change or replug.
- **Transfer merge is insert-only.** A member's own post-merge announce carries its true upstream and can beat the demoted head's transfer; the transfer must not flatten that edge onto the demoted head or a later departure prunes the wrong subtree.
- **Reports forward across demotion.** Head propagation is hop-by-hop, so a just-demoted head can be addressed under the old regime; it forwards the report to the head it now holds (the announce path self-heals on head change; the report path has no retry).
- **Head radio-slot lifecycle.** The held head is a unicast target that's usually not an adjacent HELLO peer: adoption claims its ESP-NOW slot, swap/ring-latch/INPUT-loss release it, with the same keep-slot guards as jack peers (2-node ring one-cable drop keeps wireless alive).

## Deliberate style exceptions
The three new PktType values keep the enum's existing k-style rather than going UPPER_CASE — renaming 3 of 23 values would leave the enum mixed; a whole-enum sweep should be its own change. One read of legacy `daisyChainedByPort_` mirrors the identical loop in `releaseHelloPeer` beside it.

## Validation
Native 353/353 (16 roster tests, each written failing-first), CLI 86/86, `esp32-s3_pdn_release` builds. Hardware-unvalidated; the bench sweep worth doing before an event is chain -> ring -> break-one-cable -> shootout, which exercises the ring-break suppression end to end.

## Review round 2 (post-open six-lens rerun)

Two commits landed after opening, from rerunning the review lenses against the tip:

- **67a2a43** — three delivery-reliability fixes in the report path, each with a falsified test: context-retry cancels now run before the held-head keep guard in `releaseHelloPeer` (a dying INPUT peer that IS the head could otherwise resurrect its just-released radio slot via a pending retry — permanent 16-slot-table leak in the plain 2-device chain); the DisconnectReport channel is KEEP_DISTINCT (reports are distinct events; back-to-back reports to one head were superseding each other's retry slots); and an undelivered report now chases a head change the way announces already did, instead of being cancelled with the old head's slot. Plus a fence refusing to forward a report back to its sender (head-transient ping-pong).
- **267f928** — `getChainMembers` gates on `!isRosterAuthority()` instead of a hand-written inverse; LOG_W on the two trust-boundary drops; comment accuracy fixes.

Known residual, consistent with the no-ack design: pending-report tracking is single-slot, so of multiple in-flight reports only the newest re-sends on a head change; an older one whose first transmit also failed stays in the documented residual class (needs two first-transmit failures plus a head change inside the ~1.5s retry window).

### DIFF
--- a/lib/core/src/device/remote-device-coordinator.cpp
+++ b/lib/core/src/device/remote-device-coordinator.cpp
@@ -173,8 +173,8 @@ void RemoteDeviceCoordinator::initialize(WirelessManager* wirelessManager, Seria
         // SEND_SUCCESS on the announce IS its delivery signal (#144): only then
         // may the HELLO confirmed bit rise. A head change mid-flight makes an
         // older announce's delivery stale — the re-announce to the current head
-        // is what gates confirmed — so both the seqId and the destination must
-        // still match the announce most recently sent.
+        // is what gates confirmed — so the seqId must match the latest announce
+        // sent AND the destination must still be the head currently held.
         connectionAnnounceChannel->setOnDelivered([this](uint8_t seqId, const uint8_t* toMac) {
             if (seqId != pendingAnnounceSeqId) return;
             const uint64_t head = chainHeadState.load();
@@ -1150,8 +1150,7 @@ bool RemoteDeviceCoordinator::isRosterAuthority() const {
 
 std::vector<std::array<uint8_t, 6>> RemoteDeviceCoordinator::getChainMembers() const {
     // A RING device is its own head, so it serves its roster like a HEAD does.
-    const ChainRole role = getChainRole();
-    if (role == ChainRole::CHILD || role == ChainRole::STANDALONE) return {};
+    if (!isRosterAuthority()) return {};
     std::vector<std::array<uint8_t, 6>> members;
     members.reserve(chainRoster.size());
     for (const auto& entry : chainRoster)
@@ -1238,8 +1237,8 @@ void RemoteDeviceCoordinator::transferRosterTo(const uint8_t* newHeadMac) {
     if (chainRoster.empty() || headTransferChannel == nullptr) return;
     HeadTransferPayload transfer{};
     for (const auto& entry : chainRoster) {
-        // Cannot trip while both insert paths cap at MAX_CHAIN_MEMBERS; kept as
-        // the payload array's hard bound.
+        // Both insert paths cap at MAX_CHAIN_MEMBERS; this bounds the payload
+        // array regardless.
         if (transfer.memberCount >= MAX_CHAIN_MEMBERS) break;
         memcpy(transfer.memberMacs[transfer.memberCount], entry.first.data(), 6);
         transfer.memberCount++;
@@ -1276,7 +1275,10 @@ void RemoteDeviceCoordinator::onDisconnectReport(const uint8_t* fromMac,
     // Trust boundary: no report may name this device. A member that lost the
     // link to its own head sends nothing, so a self-referential report is
     // bogus — and honoring it would erase the whole roster via the subtree walk.
-    if (memcmp(report.disconnectedMac, selfMac.data(), 6) == 0) return;
+    if (memcmp(report.disconnectedMac, selfMac.data(), 6) == 0) {
+        LOG_W("RDC", "dropping self-referential disconnect report");
+        return;
+    }
     if (!isRosterAuthority()) {
         // Head propagation is hop-by-hop, so a just-demoted head can still be
         // addressed under the old regime; forward the edit to the head now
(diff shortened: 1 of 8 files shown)


lib/core/src/device/remote-device-coordinator.cpp
```diff
@@ -1027,3 +1135,195 @@ void RemoteDeviceCoordinator::maybeFireChainRoleChange() {
...
+    }
+    uint8_t headMac[6];
+    unpackMac(headMac48, headMac);
+    ConnectionAnnouncePayload announce{};
+    memcpy(announce.upstreamMac, upstream->peer().data(), 6);
+    pendingAnnounceSeqId = connectionAnnounceChannel->sendReliable(headMac, announce);
+}
+
+void RemoteDeviceCoordinator::reportDownstreamLoss(const uint8_t* lostMac) {
+    const uint64_t headMac48 = chainHeadState.load() & HEAD_MAC_MASK;
```
cyon1c, line 1199: I've seen this a few times, what is the 48 meant to denote on these mac values?

lib/core/src/device/remote-device-coordinator.cpp
```diff
@@ -1027,3 +1135,195 @@ void RemoteDeviceCoordinator::maybeFireChainRoleChange() {
...
+    if (disconnectReportChannel == nullptr) return;
+    DisconnectReportPayload report{};
+    memcpy(report.disconnectedMac, lostMac, 6);
+    pendingReportSeqId = disconnectReportChannel->sendReliable(headMac, report);
+    memcpy(pendingReportMac.data(), lostMac, 6);
+}
+
+void RemoteDeviceCoordinator::removeChainMemberSubtree(const uint8_t* mac) {
+    // Everything downstream of the departed MAC is unreachable too: walk the
+    // upstream links to a fixpoint. `gone` doubles as the BFS frontier.
```
cyon1c: BFS frontier?

cyon1c (approved):

lib/core/src/device/remote-device-coordinator.cpp
tire-fire: 48 is the number of bits in a MAC (6 bytes). I packed it into a uint64_t so that it would be atomic for any of the writes from the serial device. Since some things happen async it will never be a partial read against the head MAC address.

lib/core/src/device/remote-device-coordinator.cpp
tire-fire: The transfer payload is just bare MACs, so the receiving head can't know the transferred members' order. It records them all as downstream of the demoted head, which is where the BFS comes from to clean up all that.
We should probably change this to be a list with (member, upstream) pairs instead of bare MACs, and handle cleanup as a separate function.

cyon1c (approved):

Issue #166, opened 2026-07-01
game: slim Quickdraw constructor; move provider injection to caller


cyon1c:
`Quickdraw` constructor is currently a ~70-line wiring harness mixing manager creation, transport initialization, callback cross-wiring, and RDC provider injection. Slim it to its core responsibilities.

**Scope**
- `Quickdraw` holds `GameContext` and owns lifetimes of its members
- `populateStateMap()` calls `QuickdrawStateBuilder::build()`
- `onStateLoop()` calls `chainManager->sync()`, `shootoutManager->sync()`, emits stats log line
- `setUserIdProvider` / `setRoleProvider` move out of constructor to `main.cpp` or a `GameSession` factory
- `onDirectPeerChange` fan-out to `ChainManager` and `ShootoutManager` replaced by their RDC callback subscriptions

**Blocked by:** #163, #167

**Reference**
[game-layer-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/145) — Quickdraw after redesign section

tire-fire:
Closed by #235

Issue #162, opened 2026-07-01
rdc: role-change context re-broadcast; half-open connection gate


cyon1c:
Two behavioral additions that plug into the context exchange layer.

**Scope**
- **Role-change re-broadcast**: register via `Player::setOnRoleChanged` to call `rdc->resendContext()` on any hunter/bounty role flip. RDC queues a fresh `PdnConnectionContext` send on both connected jack channels. Without this, a peer continues acting on a stale role until the next connection event
- **Half-open connection gate**: match initiation is only permitted when the opponent-jack peer is in `Connected` state with the duel-triggering role. Replaces `mutualOpponentIsHunter()` / PeerGraph walk with a simple local state check — same guarantee, no graph

**Blocked by:** #157

**Reference**
[rdc-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/144) — Edge case checklist, "Handled differently in new design" section

tire-fire:
Landed on `comms-rewrite` as 3f76288 (#225). Closing manually — the PR targeted `comms-rewrite`, so the `Closes` keyword did not fire.

PR #216, opened 2026-07-12
game: ConnectState per-jack connection callbacks (#165)

### LINKED ISSUE #165: game: refactor ConnectState to per-jack callback pattern

tire-fire:
Closes #165.

The issue text predates the clean-base rebuild. `onDirectPeerChange` and `roleProvider_` never existed on comms-rewrite, so there was nothing to remove.

## Answering the review on state.hpp

`onJackEvent` is off base `State`. It's gone, along with the `Quickdraw` dispatcher that needed it. `ConnectState` subscribes itself to the RDC jack observer on mount and releases it on dismount, so nothing has to ask what type the current state is. `TypedState`'s mount and dismount are final, so I added `afterMount` and `beforeDismount` to `TypedState` and `ConnectState` uses those.

That meant not every quickdraw state has to inherit `ConnectState`. The ones that don't watch jacks stay on `TypedState`.

## Shape

One delivery path: `setOnJackChange(std::function<void(SerialIdentifier, const JackConnectionState&)>)`, where `JackConnectionState` is `{connected, optional<ConnectionContext>}`. Register once, in the constructor or in `onStateMounted`. The mount replay runs after both and the slot survives a dismount, so one registration covers every tenure.

Two deviations from the #145 sketch. It's `set...` because the styleguide reserves the `on` prefix for hooks. And the context carries opaque profile bytes plus a `peerType` discriminator rather than `context->player.gameRole`, which keeps `PlayerProfile` out of core; the game layer memcpys and reads the field. That follows your "core delivers, game layer interprets" over the sketch above it. Say if you meant the sketch.

The RDC now holds the peer's kind, chainRole and profile per jack. The kind comes from the context channel that decoded the profile, so the kind and the bytes it describes always come from the same frame. `getPeerDeviceType` reads it when HELLO owns the jack, the same split `getPortStatus` already makes, and reads UNKNOWN until the exchange completes.

## Where to look

The peer kind was HELLO-sourced in an earlier commit on this branch. The two sources can disagree: the `deviceType` byte arrives over serial, the profile over ESP-NOW, and nothing cross-checks them, so a consumer casting the bytes could be handed the wrong struct. Taking both from the channel closed that, and it also deleted a write-ordering rule against the link machine's teardown-on-swap.

The RDC has one jack-observer slot and the mounted state owns it. That works because StateMachine dismounts before mounting, and that ordering is the entire guarantee. A second ConnectState mounted without it would leave the first deaf. The `subscribed` flag records that a state registered, not that it still holds the slot.

Inside a disconnect callback the polled surface hasn't caught up. `getPortStatus` and `isConnected()` still read CONNECTED for that jack.

None of this fires in production. HELLO has no caller under `src/`, so the edges and the replay are both dead until #159/#160. No hardware validation claimed.

## Two gaps in the surrounding plan

`setSelfPlayerProfile` has no caller anywhere in the tree, so with HELLO on, two PDNs would exchange 28 zero bytes and every consumer of this payload would read zeros. #162's `resendContext()` on a role flip needs a source for it and doesn't name one. #166's `setUserIdProvider` / `setRoleProvider` bullet names symbols that don't exist on this base, the same way two of #165's bullets did.

`native_asan` cannot complete on this branch. `Match::Match` memcpys a fixed 36 bytes from the caller's pointer (`match.cpp:9`) and `match-tests.hpp:42` passes the 8-byte literal `"match-1"`, so ASan aborts at `MatchTestSuite.jsonContainsWinnerFlag`. Identical code on comms-rewrite, and it aborts with this branch's changes stashed too, but ASan only trips when the globals layout puts a redzone after that literal and editing `tests.cpp` perturbs that layout. Filtered ASan over the ConnectState and RDC suites is clean.

## Deferred

- `getPeerMac` and `getPeerUserId` still read the handshake table. #159.
- `setOnContextReceived` now pushes the same facts `onJackChange` carries, and has no consumer under `src/`. Left in place because it's #157's surface and may be intended API. Say if it should go.
- The disconnect drops the departed MAC, which the RDC still holds at that instant. ShootoutManager's abort is keyed on it. Needs `getPeerMac` first.
- `Idle::transitionToSymbol` fires on an FDN on either jack while `SymbolState` only reads the OUTPUT jack's MAC, so an INPUT-jack FDN bounces Idle to Symbol and back through a 500ms blocking spin. Live today, pre-existing, own PR.
- `~Quickdraw`'s handler-clear list duplicates the constructor's seven registrations. Pre-existing, own PR.


lib/core/include/state/state.hpp
```diff
@@ -110,6 +115,10 @@ class State : public StateLifecycle {
 
     virtual bool isTerminalState() { return false; }
 
+    /// Per-jack connect/disconnect delivery from the app's dispatcher; the
+    /// default ignores it so only states that observe jacks need to override.
+    virtual void onJackEvent(SerialIdentifier jack, const JackConnectionState& state) {}
```
cyon1c: Why is this applied to state and not connected-state?

tire-fire: Quickdraw only ever holds a plain State* and since we don't compile with RTTI to do introspection we can't determine the type, so I add the blank onJackEvent to the base and let ConnectedState implement it. `isTerminalState` also does this if you hate this style, I can come up with a different way to do this.

cyon1c: I think this takes just a little bit of rethinking - I do believe this belongs on connectstate and not on the base state.

cyon1c: I'm wondering if we can just have every quickdraw state uniformly inherit ConnectState and get around it that way?

tire-fire: ConnectState registers itself on mount using new afterMount and beforeDismount on TypedState, so onJackEvent is gone from State.

lib/core/include/state/state.hpp
```diff
@@ -165,16 +170,27 @@ class TypedState : public State {
...
+    virtual void beforeDismount(DeviceT* device) {}
+
 private:
     // Private final bridge — casts once and forwards to the typed user API above.
     // State implementers cannot override or call these.
     void mount(Device* device) final {
-        onStateMounted(static_cast<DeviceT*>(device));
+        DeviceT* typed = static_cast<DeviceT*>(device);
+        onStateMounted(typed);
+        afterMount(typed);
```
cyon1c, line 186: love.

cyon1c (approved):

cyon1c (approved):

PR #221, opened 2026-07-26
Crash Log Reporting - With some extras on the side

### LINKED ISSUE #133: Log Crashes
We need a framework for capturing crash logs and transmitting them to the server.

cyon1c:
This PR sets up the PDN to allow for crashlogs to be captured for later inspection.

There are 3 possible routes for extracting crash reports:

Serial: With the serial monitor open, sending `CRASHLOG` to the serial monitor at any point will print out the last 10 crashes on the device.
ESP-NOW: On boot, any unbroadcasted crashes are sent over broadcast and marked as sent.
HTTP: Unimplemented at the moment as we will need to figure out HTTPS/TLS on device before testing against our dev server.

In pursuit of setting this up, there were a few side effect changes.

1) Partitions.csv & sdkconfig.defaults - this sets up a custom partition scheme for the RAM on the ESP32, setting aside part of the RAM for storing crash logs that can be retrieved after reboot.
2) Updates to the storage driver - the storage driver now accepts an array of namespaces and creates a prefs object for each one, instead of having only 1 (previously only match storage).
3) Quick rename of the esp-now driver from EspNowManager to EspNowDriver (wasn't required but it was bugging me!)

Here is an example of a crash log over serial from forcing an abort in idle state on connection:

```
=== CRASH LOG ===
Device crash count: 1 (0 pending transmission)
[#1][sent   ] reason=4  PANIC        | ts=1871ms | pc=0x4037E49D | ec=0  IllegalInstruction | task=loopTask
=================

(crash log dump complete)
```

**NOTE**: On first deploy of this firmware, you will need to ensure that you run a full clean erase of the firmware. This can be done with the multi_flash script while specifying the --erase flag, or compiling and deploying the production version of the firmware.


tire-fire:
`abort()` is back in `Idle::onStateLoop` at head, `src/pdn/game/quickdraw-states/idle-state.cpp:76-78`. 2ca3e77d removed it and a5dbdcba re-added the same four lines, looks like a rebase artifact.

The namespace argument on `StorageInterface` is what pulls match-manager, hacked-players-manager, perf-main, the device mock and both test files into this PR. Every call site passes one constant, and `CrashLogger` holds a concrete `Esp32S3PrefsDriver*` rather than the interface, so it never uses the widened signature. A second `Preferences` inside `CrashLogger` would cover the same need without the interface change. One side effect either way: `findPrefs` returns null for an unregistered namespace and every method then silently no-ops, so a typo turns into vanished writes with no error.

Nothing registers a `setPacketHandler` for `kCrashLog`, so those frames are dropped on receipt. `EspNowDriver::sendData` also discards `SendFrontPkt`'s return and always gives back 0, so `sentSeq` advances on queue rather than delivery. A record whose send failed gets marked sent permanently, which is the opposite of what the `LOG_E` on the else branch promises.

Two smaller things. `sdkconfig.defaults` isn't read unless `custom_sdkconfig` is set on the env or board, and those three values are already `y` in the prebuilt arduino libs, so the `coredump` partition is doing all the work on its own. And I don't think the full-erase note is needed, `nvs` lands at the same offset and size as `default_8MB.csv` so existing keys should resolve fine. The old `app1` region does end up under `coredump`, but the CRC check rejects the stale bytes and erases them.

cyon1c:
@tire-fire Addressed your comments, sans the ESP-NOW stuff. This isn't an actual implementation of the esp-now route but i wanted to get _something_ in. Once we figure out our pdn <-> fdn network protocol, we can wire this up to that.

tire-fire (approved):

Issue #218, opened 2026-07-14
scale: big chains and rings end to end — broadcast bracket, roster cap, 115200 baud


tire-fire:
Two chains near the roster cap linking together (raised on #158's review) exposed that our scale ceilings are scattered and partly obsolete. This lifts them as one coherent work-item instead of nibbling per subsystem.

**Current ceilings, verified against the code:**
- `MAX_CHAIN_MEMBERS = 18` caps the head roster. On overflow, announces and transfer entries drop (logged), but the dropped member's HELLO confirmed bit still rises — SEND_SUCCESS is the delivery signal and the radio acks a frame the head discarded — so devices past the cap display confirmed yet never enter the shootout bracket. The 18 came from fitting `HeadTransferPayload` in one ESP-NOW v1 frame (250B). The fleet is single-firmware on IDF 5.5, so every link negotiates ESP-NOW v2: one frame now carries ~240 MACs, and the framing rationale is gone.
- Shootout bracket fan-out unicasts the full bracket to every member; the ESP-NOW peer table is hardware-capped at 20 slots, so sends start failing somewhere in the high teens once neighbor/head slots are counted. (The older OOM concern does not hold in the current code: the transient is ~6N² payload bytes ≈ 15KB at N=50.)
- `RING_EVIDENCE_TIMEOUT_MS = 500` covers ~16 hops of head-claim propagation at today's per-hop latency: 20ms HELLO cadence + ~9.9ms frame time at 19200 baud.

**Plan:**
1. **Broadcast the bracket** (one permanent broadcast peer slot), keep per-member acks unicast — members already hold the coordinator's slot. One frame per retry round replaces N unicasts; a v2 frame carries a ~240-member bracket. This supersedes the parked LRU-eviction approach to the peer table.
2. **Raise `MAX_CHAIN_MEMBERS`** (single v2 HeadTransfer frame bounds it at ~240; pick to the event envelope, e.g. 64) and settle the overflow contract — at minimum the confirmed-bit lie past the cap should be named or fixed, not silent.
[...]

PR #223, opened 2026-07-27
test: RDC peer-swap slot ordering and mid-frame drop coverage

### LINKED ISSUE #170: test: port RDC and SerialFrameParser tests from PR
Port `rdc-tests.hpp` and `serial-frame-parser-tests.hpp` from the PR to the new design. The design doc explicitly requires the three existing RDC tests from the PR be preserved alongside the rewrite.

**Scope**
- Port `serial-frame-parser-tests.hpp` — frame encode/decode, reset behavior, HELLO opcode only
- Port the three existing RDC tests from the PR that the design doc mandates
- Update test harness to call `serviceConnectivity()` directly via `setExternalConnectivityTask(true)`
- Adapt `device-mock.hpp` as needed for new RDC API surface

**Blocked by:** #155

**Reference**
[rdc-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/144) — Edge case checklist, one-way link section ("three existing tests from the PR should be ported")

tire-fire:
Closes #170.

Most of this issue was already satisfied. `serial-frame-parser-tests.hpp`, `rdc-tests.hpp` and `rdc-hello-tests.hpp` are all present, the harness already drives `serviceConnectivity()` through `setExternalConnectivityTask(true)`, and `device-mock.hpp` already carries the new RDC surface. Rather than re-porting what exists, this adds the two mandated behaviours that had no coverage.

## What was actually missing

`peerSwapReleasesOldSlotThenAdoptsNew` — a HELLO from a different source MAC on a live jack must release the old peer's radio slot before adopting the new one. The ordering is the point: `HelloLinkMachine::onHelloReceived` tears the link down to Idle first so the old parser and chain state are gone before the same frame opens the new link.

`jackDeathDropsHalfReadFrame` — a link dying mid-frame must leave no partial frame behind for the next peer to finish. `HelloIdleState`'s mount is the single teardown seam and requests the parser reset there.

Both were falsified by reverting their mechanism and confirming failure.

## Where to look

The test baseline is 364 → 366. An earlier draft of this description said 378 → 380; that number came from counting against another in-flight branch, not this base.

## Not fixed here

This work surfaced a real defect in `SerialFrameParser`: it never resyncs after a CRC failure, so a truncated frame keeps consuming into the next frame and swallows a good HELLO whole. A timeout cannot recover it, because the driver drains the whole RX FIFO in one loop and the two frames arrive with no gap. That is pre-existing and it invalidates the drop-rate premise behind raising the baud rate, so it is fixed in #218 alongside that work rather than here.

No hardware validation was performed.

### DIFF
--- a/test/test_core/rdc-hello-tests.hpp
+++ b/test/test_core/rdc-hello-tests.hpp
@@ -127,6 +127,16 @@ public:
     SerialIdentifier lastDisconnectJack = SerialIdentifier::OUTPUT_JACK;
 };
 
+// Drives a jack Idle -> Connecting -> Connected. Each transition commits on a
+// sync() tick, and the context exchange only advances a Connecting link.
+inline void connectJack(RDCHelloTests* suite, NativeSerialDriver& jack,
+                        SerialIdentifier id, const std::vector<uint8_t>& firstHello) {
+    suite->deliverHello(jack, firstHello);
+    suite->rdc.sync(&suite->device);
+    suite->rdc.onContextExchangeComplete(id);
+    suite->rdc.sync(&suite->device);
+}
+
 // A looped-back own HELLO or an all-zero-source (open-jack) HELLO must not open a
 // link: on real hardware the output TX pin bleeds back through the TRS contacts,
 // and treating that as a peer would keep a dead cable "alive". A genuine distinct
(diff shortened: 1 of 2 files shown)


cyon1c (approved):

cyon1c:
Test only prs can just get merged i think moving forward. We'll figure out how to deal with that.

PR #226, opened 2026-07-27
scale: broadcast ring fan-out, roster cap 64, parser resync

### LINKED ISSUE #218: scale: big chains and rings end to end — broadcast bracket, roster cap, 115200 baud

tire-fire:
Steps 1-3 of #218, plus the serial parser resync the baud step turns out to
depend on. No timing constant changes in this PR.

## Broadcast bracket

The shootout ring fan-out was one unicast per member. The ESP-NOW peer table holds
20 entries, so a ring wider than that could not be addressed at all. `sendToPeers`
now emits one broadcast frame (the broadcast peer is registered once at radio init),
and the retry path emits one frame per round instead of one per pending peer. Acks
stay unicast, one per member, and per-member pending-ack tracking is unchanged.
This supersedes the parked LRU peer-table eviction patch.

I widened this past the bracket itself. All five ring fan-outs — CONFIRM, BRACKET,
MATCH_START, MATCH_RESULT, TOURNAMENT_END, plus PEER_LOST/ABORT — already route
through `sendToPeers`/`sendReliablyToPeers`, so broadcasting there is a smaller
change than special-casing BRACKET, and special-casing BRACKET would leave the
identical peer-table wall at MATCH_START one packet later.

The fragile part is the consequence, not the send. A broadcast lands on every ring
in radio range, so each receive handler now has to reject another tournament's
commands — that filtering used to come free from the destination MAC:

- BRACKET: our own MAC must be in the offered roster. Rejected before the ack, or a
  neighbouring coordinator counts us as one of its members.
- MATCH_START / MATCH_RESULT / TOURNAMENT_END: the MACs they name must be in our
  bracket. MATCH_RESULT's ack moved after this check.
- PEER_LOST: the named MAC must be a ring member.
- ABORT: carries no MACs, so `onAbortReceived` now takes `fromMac`. Without it a
  neighbouring ring's abort tears down every tournament in range.

Two places to look hard at. First, MATCH_RESULT and TOURNAMENT_END gate on
`bracket`, so one arriving ahead of BRACKET is now dropped and re-sent rather than
applied against an empty bracket — the retry budget (3 rounds, 100/200/400 ms)
against the 5 s BRACKET_REVEAL window is what makes that safe. Second, a locally
observed jack loss keeps its old unfiltered path (`applyPeerLoss`); only the remote
PEER_LOST is filtered, because during PROPOSAL a not-yet-confirmed peer is in
neither `confirmedSet` nor an already-broken loop set.

The first commit is a mechanical rename with no behaviour change. `check_style.py`
flags any added line carrying a legacy trailing-underscore member, so touching
shootout-manager at all forced it; separating it keeps the second commit readable.

## Roster cap

MAX_CHAIN_MEMBERS 18 → 64. The 18 came from fitting HeadTransferPayload into one
v1 250-byte frame, but the thing that actually bound the payload was
`DataPktHdr::pktLen` being uint8_t. It is uint16_t now, and MAX_PKT_DATA_SIZE
derives from ESP_NOW_MAX_DATA_LEN_V2 (1470) instead of the length field. A
64-member HeadTransfer is 770 bytes, a 64-member bracket 387. `kMaxBracketSize`
became `MAX_BRACKET_SIZE = MAX_CHAIN_MEMBERS`; left at 32 it would have rejected a
large ring's BRACKET at parse.

Overflow contract — named, not fixed. Past the cap the head drops the announce and
the member cannot learn that: SEND_SUCCESS on the announce is the only delivery
signal the no-ack design has, and the radio acks a frame the head then discards, so
the member's HELLO confirmed bit rises while it is absent from the roster. Closing
that needs an admission reply the no-ack decision excludes. So: the cap now sits
past any real chain, the drop logs at error level naming the consequence, and
`announceOverCapIsDroppedNotAdmitted` pins the head-side half (the member-side bit
is not observable from the head). The head-transfer overflow is a different case and
self-heals — the head change that triggers the transfer clears each member's
confirmed bit and makes it re-announce.

## Receive-side pktLen check

Widening `pktLen` also widened what a wrong value can ask for. It comes off the
wire, it is what bounds the copy handed to the packet handlers, and nothing
compared it against what the radio actually delivered. The recv callback now
drops a frame whose declared length disagrees with `data_len`, or that claims to
be shorter than the header. That second case underflowed `handleSinglePacket`'s
length subtraction to roughly SIZE_MAX. `sendData` is the only writer and sets
the two equal, so nothing legitimate is rejected.

## Parser resync after a CRC failure

Worth reading before the baud discussion below, because it invalidates the
premise the issue raises the rate on. #218 assumes 115200 buys a higher drop rate
at unchanged integrity. That only holds if a dropped byte costs one frame, and it
did not. On a CRC mismatch `SerialFrameParser` reset to scan-for-sync and threw
away every byte it had consumed. A truncated frame keeps reading into the next
frame's bytes, fails CRC on those, and takes the good HELLO behind it down with
it. Two frames per drop.

The mid-frame timeout cannot rescue that. The ESP32 driver drains the whole RX
FIFO in one `while (Serial1.available() > 0)` loop and hands each byte straight
to the parser, so the short frame and the good one behind it can arrive with no
gap at all. At a 20 ms cadence against a 100 ms silent-link timeout that is a
real bite out of the margin, taken exactly when the drop rate rises.

The parser now replays the buffered payload plus the CRC bytes back through
scan-for-sync instead of discarding them, so an embedded 0xAA 0x55 restarts
framing at the true frame start. The disproven preamble and opcode are not
replayed, which is also what bounds the re-entry: the replay is 16 bytes and a
whole frame is 19, so it can never reach a second CRC comparison. A trailing lone
0xAA leaves the parser in GOT_AA and pairs with whatever byte arrives next, which
is the common shape when the byte that went missing was mid-payload.

`midFrameTimeoutResets` advances 60 ms, a gap the 20 ms cadence never produces,
so it never covered any of this. The two new tests advance one cadence: a
truncated frame then a good one, and a frame with a payload byte dropped then a
good one. Both expect exactly one decode and both fail if the replay is removed.

## BAUDRATE type

uint16_t → uint32_t. 115200 wrapped to 49664 in 16 bits, which would have silently
configured the UART for a different speed. Live and latent independently of whether
the rate changes. A `static_assert` now fails the build if the type narrows again.

## Deferred to the hardware sweep (step 4 — not done here)

Three values are deliberately unchanged. Each is a timing constant on
hardware-proven code and each needs bench evidence this run cannot produce. The
parser resync is a prerequisite for that sweep rather than a part of it: run the
sweep without it and every dropped byte reads as two lost HELLOs, which lands on
the TRS contacts in the results instead of on the parser.

- **BAUDRATE 19200 → 115200.** A framed HELLO is 19 bytes (2 preamble + 1 opcode +
  14 payload + 2 CRC). At 19200 8N1 that is 190 bits / 19200 = **9.9 ms**; at 115200
  it is **1.65 ms**. Hop propagation stops being frame-bound and becomes
  cadence-bound. Needs TRS-contact reliability at 115200 measured, not assumed.
- **HELLO_CADENCE_MS 20.** At 115200 a hop costs cadence + 1.65 ms, so a 10 ms
  cadence puts a hop at ~11.65 ms.
- **RING_EVIDENCE_TIMEOUT_MS 500.** Today 500 / (20 + 9.9) ≈ **16 hops**, which is
  what its comment claims. At a 10 ms cadence and 1.65 ms frame time,
  500 / 11.65 ≈ **43 hops** — the same 500 ms already covers a 64-device ring, so
  this one likely needs no change at all. Whichever cadence is picked,
  HELLO_SILENT_LINK_MS must stay several cadences (100 ms is 5 × 20 ms today; a
  10 ms cadence wants ~50 ms, not less).

The sweep they wait on: TRS contacts at 115200, long-chain head propagation, a
big-ring close, and one full broadcast-bracket round. **Steps 1-3 are done. Step 4
is not.**

Two ceilings also stay put, because they are peer-table pressure rather than framing
and lowering that pressure wants the same hardware run: `kMaxChainPeersPerPort` is
18 per port (each daisy-chained peer takes an ESP-NOW slot via
`RemoteDeviceCoordinator::registerPeer`), and the shootout's loop-member set — the
set the bracket is built from — is still assembled from those per-port lists. A
64-member head roster is now representable; a 64-device *ring* is not yet reachable
through discovery.

Closes #218

### DIFF
--- a/lib/core/include/device/drivers/peer-comms-types.hpp
+++ b/lib/core/include/device/drivers/peer-comms-types.hpp
@@ -122,9 +122,9 @@ struct ShootoutAckPayload
 // Point-to-point ReliableChannel payloads; the chain head is the sole consumer.
 // seqId is stamped by the channel on send.
 
-// Head-roster capacity, sized to the event envelope rather than to a frame:
-// a single ESP-NOW v2 HeadTransfer frame holds well over a hundred (member,
-// upstream) pairs, so the old 18 (one v1 250-byte frame) no longer binds.
+// Head-roster capacity, sized to the event envelope rather than to a frame: a
+// single ESP-NOW v2 HeadTransfer frame holds well over a hundred (member,
+// upstream) pairs, so framing is not what bounds this.
 //
 // Overflow contract: past this cap the head drops the announce, and the member
 // cannot learn that. SEND_SUCCESS on the announce is the only delivery signal
--- a/lib/core/include/device/drivers/peer-comms-types.hpp
+++ b/lib/core/include/device/drivers/peer-comms-types.hpp
@@ -64,8 +64,10 @@ struct FdnConnectionContext {
 
 struct DataPktHdr
 {
-    //Total packet length including header
-    uint8_t pktLen;
+    // Total packet length including header. Wide enough for a full ESP-NOW v2
+    // frame: the fleet is single-firmware on IDF 5.5, so every link negotiates
+    // v2 and a payload can exceed 255 bytes.
+    uint16_t pktLen;
     PktType packetType;
 } __attribute__((packed));
 
@@ -120,7 +122,17 @@ struct ShootoutAckPayload
 // Point-to-point ReliableChannel payloads; the chain head is the sole consumer.
 // seqId is stamped by the channel on send.
 
-constexpr uint8_t MAX_CHAIN_MEMBERS = 18;
+// Head-roster capacity, sized to the event envelope rather than to a frame:
+// a single ESP-NOW v2 HeadTransfer frame holds well over a hundred (member,
+// upstream) pairs, so the old 18 (one v1 250-byte frame) no longer binds.
+//
+// Overflow contract: past this cap the head drops the announce, and the member
+// cannot learn that. SEND_SUCCESS on the announce is the only delivery signal
+// the no-ack design has, and the radio acks a frame the head then discards, so
+// the member's HELLO confirmed bit rises while it is absent from the roster.
+// The cap is set past any real chain so the divergence is unreachable in
+// practice; closing it would need an admission reply the design excludes.
+constexpr uint8_t MAX_CHAIN_MEMBERS = 64;
 
 // Sent by a newly joined member directly to the head it read from its upstream
 // neighbour's HELLO. upstreamMac names the sender's direct upstream (INPUT) peer.
@@ -140,10 +152,10 @@ struct DisconnectReportPayload {
 // ever travels, and it is a single unicast. Entries are (member, upstream)
 // pairs — memberMacs[i]'s recorded direct upstream is upstreamMacs[i] — so the
 // receiver keeps the chain's true order instead of flattening every
-// transferred member onto the demoted head. At MAX_CHAIN_MEMBERS=18 this is
-// 218 bytes. The ceiling is the driver's reliable-payload budget
-// (MAX_PKT_DATA_SIZE, the uint8_t length header, ~253 bytes), asserted where
-// the driver includes this header; raising MAX_CHAIN_MEMBERS past ~20 trips it.
+// transferred member onto the demoted head. At MAX_CHAIN_MEMBERS=64 this is
+// 770 bytes. The ceiling is the driver's reliable-payload budget
+// (MAX_PKT_DATA_SIZE, one ESP-NOW v2 frame), asserted where the driver includes
+// this header.
 struct HeadTransferPayload {
     uint8_t seqId;
     uint8_t memberCount;
--- a/lib/core/include/device/serial-frame-parser.hpp
+++ b/lib/core/include/device/serial-frame-parser.hpp
@@ -98,6 +98,8 @@ private:
 
     void resetParser();
     void checkTimeout();
+    void consumeByte(uint8_t b);
+    void resyncFromConsumedBytes();
     static size_t opcodePayloadLen(uint8_t opcode);
     static unsigned long nowMs();
 };
--- a/lib/core/include/protocol-constants.hpp
+++ b/lib/core/include/protocol-constants.hpp
@@ -3,7 +3,11 @@
 #include <string>
 
 // Serial communication speed shared by all serial drivers and the handshake protocol.
-constexpr uint16_t BAUDRATE = 19200;
+// uint32_t, not uint16_t: common rates above 65535 (115200 among them) wrap in 16
+// bits and would silently configure the UART for a different speed.
+constexpr uint32_t BAUDRATE = 19200;
+static_assert(static_cast<decltype(BAUDRATE)>(115200) == 115200,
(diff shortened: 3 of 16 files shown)


cyon1c (approved):
Love to see the style cleanup along with the code changes!

PR #228, opened 2026-07-28
game: split quickdraw state graph into sub-graph builders

### LINKED ISSUE #167: game: split populateStateMap into sub-graph builders and QuickdrawStateBuilder orchestrator
`populateStateMap()` is ~335 lines creating 21 states, defining all transitions, and wiring three distinct game modes in a single flat list. Split into three focused builders and a thin orchestrator for cross-graph transitions.

**Scope**
- `DuelGraph::build(sm, ctx)` — Sleep -> Awaken -> Idle -> DuelCountdown -> Duel -> DuelPushed / DuelResultReceived -> DuelResult -> DuelOutcome(win/lose) -> UploadMatches -> Sleep
- `ShootoutGraph::build(sm, ctx)` — ShootoutProposal -> BracketReveal -> Spectator / Eliminated -> FinalStandings -> Idle + Aborted (reachable from any shootout state via `ShootoutAwareState::tickAbortGuard()`, not manual edges)
- `SymbolGraph::build(sm, ctx)` — Symbol -> SymbolMatched -> Symbol or Idle
- `QuickdrawStateBuilder::wire(sm, ctx)` — cross-graph transitions only: Idle->ShootoutProposal, DuelResult->ShootoutSpectator/Eliminated, Duel->ShootoutSpectator/Eliminated
- Each builder is a free function or small class; creates its states, adds internal transitions, returns the state list for registration

**Blocked by:** #163, #164, #165

**Reference**
[game-layer-redesign](https://github.com/alleycatassetacquisitions/pdn/discussions/145) — Three sub-graphs, one thin orchestrator section

tire-fire:
Closes #167.

`Quickdraw::populateStateMap` drops from 313 lines to 12: it fills the `GameContext` and calls `QuickdrawStateBuilder::build`. The new `quickdraw-state-builder.{hpp,cpp}` holds `DuelGraph`, `ShootoutGraph` and `SymbolGraph` — each allocating its states, wiring the edges leaving them, and appending them — plus the orchestrator that owns the cross-graph endpoints and the single shared abort predicate. Ownership is unchanged: the graph structs are stack temporaries holding non-owning pointers, `stateMap` owns all 21 states, `~StateMachine` deletes them.

## Where to look: this deviates from the issue's wording, deliberately

The issue says `QuickdrawStateBuilder::wire` adds the cross-graph transitions. Doing that literally changes behaviour. A transition list is a priority list — `State::checkTransitions` returns the first condition that holds — so appending cross-graph edges after each graph's internal ones silently demotes them. `Idle` interleaves them (`shootoutProposal` at slot 1, `shootoutAborted` at 4, `symbol` at 5), as do `DuelResult` and `SymbolMatched`. So the orchestrator owns the cross-graph *endpoints and the shared predicate*, and each graph applies them in place at the exact slot in its own state's list.

Demoting Idle's `shootoutProposal` edge below its `duelCountdown` edge is not hypothetical — the code carries a comment saying that turns a ring's tournament intent into a 1v1 duel.

## The coverage this refactor needed and did not have

The state graph had **zero** automated coverage. The only `Quickdraw` a test builds never calls `StateMachine::initialize()`, so `populateStateMap` never ran under test: neither the core suite nor the CLI suite asserted the registration order or any of the 48 transitions. A green suite was not evidence for a refactor whose entire risk is ordering.

So this adds `quickdraw-state-graph-tests.hpp`, pinning both orders against the graph as it stood before the split:

- **registration order** — 21 ids, `PlayerRegistrationApp` at index 0. That index is what `StateMachine::initialize` mounts, and `skipToState` is index-addressed with two CLI call sites passing a literal `1`.
- **transition order** — every edge, per state, in priority order, 48 total.

Both were falsified: moving `playerRegistration` off index 0 fails only the registration test; swapping Idle's first two edges fails only the transition test. Verified independently of the tests too, by diffing per-source edge order and registration order against f74d4b4 — identical on both.

This needs one addition to the core framework: a read-only `State::getTransitions()`. Reading a transition list does not evaluate its conditions, which is what lets the graph be inspected with a null `GameContext` before any manager exists.

## Deliberately not done

Six simplification findings survived review and are left for you rather than auto-applied, all on code this PR just wrote:

- `build()` and `appendTo()` are two hand-maintained copies of the same ordering; collapsing them changes three signatures and has an ordering landmine (`playerRegistration`'s transition must be added after `duelGraph.build` or it captures a null target into a `StateTransition` that has no null guard).
- `CrossGraphTargets` duplicates fields the graph structs already hold; a missing neighbour is a silent nullptr rather than a compile error. Passing explicit parameters would make it an arity error.
- 38 of 48 transitions hand-expand the same member-pointer call shape; a small helper erases them and the local alias preamble each `wire()` needs.
- `supporterReadyState` caches a pointer the state machine already owns, forcing `build`'s return type.

One comment was corrected rather than left: `ShootoutGraph`'s doc claimed `phaseIsAborted` was "the single predicate" observing the abort. It is not — four shootout states reach Aborted through their own per-state latch, and `ShootoutSpectator`/`ShootoutEliminated` are not `ShootoutAwareState` at all.

No existing test was added to, weakened, or edited. No hardware validation was performed. `Closes #167` will not fire — the base is not the default branch.


cyon1c:
Rather than building this big coordinator that is only for the Quickdraw game specifically, can we build this in a way that we're doing state-machine swaps and keep them less coupled together?

tire-fire:
Superseded by #232, which takes the state-machine-swap approach you asked for rather than the coordinator here.

`Quickdraw` is gone and Duel/Shootout/Symbol are three apps in `AppConfig` alongside `PlayerRegistrationApp`, swapped through the same `addAppTransition` path FDN already uses. Three framework fixes were needed to make that work, one of which repairs a live FDN defect: every app mount was re-running `populateStateMap` and leaking a full state set.

The graph pinning test from this PR carries over as the equivalence proof rather than being discarded — all 48 edges, plus which 18 became cross-app hand-offs.

Closing this in favour of #232 unless you'd rather compare them side by side.

tire-fire:
Closing in favour of #232, which takes the state-machine-swap approach cyonic asked for here. The graph-pinning test this PR introduced carries over there and does the equivalence proof for the split; nothing else in this branch survives the change of approach.

#167 stays open — it is what #232 implements, and `Closes` will not fire on a non-default base.

PR #232, opened 2026-08-03
game: split quickdraw into swappable duel, shootout, symbol apps

### LINKED ISSUE #167: game: split populateStateMap into sub-graph builders and QuickdrawStateBuilder orchestrator

tire-fire:
Supersedes #228, which took the opposite approach.

Closes #167.

> Rather than building this big coordinator that is only for the Quickdraw game specifically, can we build this in a way that we're doing state-machine swaps and keep them less coupled together?

Yes, and the mechanism already existed. `Quickdraw` is deleted. Gameplay is now four apps in `AppConfig` alongside `PlayerRegistrationApp`, swapped by the same `addAppTransition` / `setActiveApp` path FDN has always used. No app holds a pointer into another. A cross-app edge names an app id and a state id, nothing more. PDN registered one app before this and now registers five, like FDN.

> can we make Awaken/sleep/idle/supporter ready into a separate app from duel? This way we can use this "hub" app as our standard base, and then duel starts at DuelCountdown.

Done, and it is the fourth app. `HubApp` holds Awaken, Idle, SupporterReady and Sleep. `DuelApp` starts at DuelCountdown. Every other app hands back to the hub. That split moved six edges from intra-app to hand-offs: Idle's launch into a duel, the four abandoned-duel returns, and the upload's exit to Sleep.

Three framework fixes made the swap possible. They are the part worth reviewing.

## Framework fix 1: app mounts leaked a whole state set, and FDN is affected today

`StateMachine::initialize()` ran `populateStateMap()` unconditionally on every mount, nothing cleared `stateMap`, and no app guarded re-entry. Probed before touching anything:

```
mount 1: populateStateMap calls=1  stateMap.size=2  current=1
mount 2: populateStateMap calls=2  stateMap.size=4  current=1
mount 3: populateStateMap calls=3  stateMap.size=6  current=1
mount 4: populateStateMap calls=4  stateMap.size=8  current=1
```

Every swap allocated a duplicate state set that was never mounted, because `currentState` stays pinned to the original slot, and never freed until `~StateMachine`. **This is live on FDN**, whose four apps swap freely, so a session accumulates a state set per swap on constrained RAM. Now populates once.

Consequence worth knowing: the 20 gameplay states used to be allocated at boot inside one `populateStateMap`. They are now allocated lazily, so DuelApp's 8 land on the first duel hand-off. Total heap is the same or lower. `new` results are pushed unchecked, so an allocation failure under fragmentation would surface as a null `currentState` mid-duel rather than at boot.

## Framework fix 2: an app transition names the state it enters

Re-entry always restarted at `stateMap[0]`, and `setActiveApp` took no target. 19 of the 24 cross-app edges enter somewhere other than the target's boot state. Every abort lands on `ShootoutAborted`, the four abandoned-duel returns and the symbol app land on `Idle`, a finished upload and a finished tournament land on `Sleep`, and a bracket outcome lands on the shootout's spectator or eliminated. Without this they would all have meant "restart the target from the top".

The target is a `StateId`, resolved by `StateMachine::findEntryState` against the target's own state map, so ids only have to be distinct within one app. An id no state carries logs and falls back to the boot state. Omitting it means the boot state, which is what all ten FDN call sites want, so no FDN site changed.

Both `Device` mount paths write the entry state before mounting rather than letting one inherit what the last swap left. A `StateMachine` driven outside an `AppConfig` never writes it and rides the unset default; `HelloLinkMachine` is the one such caller, so that default is load-bearing.

**App and intra-machine edges now share one priority list.** `checkAppTransitions` is gone; `addAppTransition` pushes into `transitions`. This is a semantic change and it was forced. Checking local edges first would demote Idle's shootout hand-off below its local edge into SupporterReady, and that ordering is what stops a closed ring's tournament being silently demoted to a 1v1 duel.

That changes ordering for any state declaring an app edge before a local one. Nine of FDN's ten app transitions were already declared last on their state, so nothing there was re-ranked. `symbolIdle` was the exception: its `MAIN_MENU` edge moved below `matchSuccess` to keep its previous effective priority. That edge is inert either way, since `transitionToMainMenuApp` is never assigned true.

## Framework fix 3: managers no longer live inside an app

Only the mounted app's `onStateLoop` runs, but `MatchManager`, `ChainDuelManager` and `ShootoutManager` are shared and were pumped from `Quickdraw::onStateLoop`. Split naively, manager sync would stop whenever the wrong app was mounted, silently.

They move to a new `GameSession`, which installs its own per-tick pump on a new `Device::setTickCallback` in its constructor and clears it in its destructor. That runs between `execDrivers()` and the mounted app's loop, the slot `Quickdraw::onStateLoop` occupied, so per-tick ordering is preserved. The slot is singular, so a second claimant would silently stop pumping the first; the doc says so.

Side effect worth knowing: the five per-state `shootoutManager->sync()` calls are gone, so tournament retries now also run while the device is in the duel app for a bracket match. That includes the MATCH_START watchdog, which previously could not fire while a coordinator was duelling. It re-announces the same match, every peer dedupes and acks, and no local state is re-primed. It is extra on-air traffic on a radio whose peer table is already the scaling ceiling.

## The abort rule

Ten states can be interrupted by a tournament abort: the hub's Idle, the five interruptible duel states, and four of the shootout states. All ten read one condition over `ShootoutManager`'s phase, so the rule has a single definition across three state maps rather than a latched copy per state.

## How equivalence is shown

`quickdraw-state-graph-tests.hpp` carries #228's pinned literals across and turns them into the proof rather than discarding them. The four gameplay apps' state maps concatenate to the pre-split registration order, and all 47 gameplay edges still leave the same state at the same list position for the same target, each hand-off resolved through to the state it actually enters. A second table pins which 24 became hand-offs, the app and entry state each names, and that no other edge did. The 48th edge, `PlayerRegistration -> AwakenSequence`, moved onto `WelcomeMessage`, its only trigger, and is pinned separately.

One deliberate divergence from the pre-split order: `Sleep` moved into the hub. Its edges are unchanged; only its flat position moved.

That file did not exist before #228. The state graph had no coverage at all, because the only `Quickdraw` any test built never called `initialize()`, so `populateStateMap` never ran under test.

## Where it is fragile

`GameSession` owns the ESP-NOW demux for eight packet types and decodes payloads whose send-side packing lives in the managers, where `RemoteDeviceCoordinator` and `ReliableTransport` each register and parse their own. Its constructor and destructor have to stay in sync by hand over those eight.

The ten abort edges are still hand-declared per state, and the test that checks them indexes a hand-maintained `(state, edge index)` table. It catches a listed state losing its edge or ceasing to honour the condition. It cannot see a state that gains one.

`skipToState(index)` still addresses by vector position, so FDN keeps three slot constants coupled by hand to its `populateStateMap` order.

The CLI `reboot` command changed behaviour. It used to land on `AwakenSequence` via `skipToState(pdn, 1)` and now lands on `FetchUserDataState`. The command's own result message already said `-> FetchUserData`, so this is almost certainly a latent fix, but no commit called it out. Simulator only; production has no reboot path.

## Not on hardware

The last sweep is the six-device chain to ring to shootout run from June, which predates this, #230 and #231. The abort path, the cable-pull teardown and the new bracket-duel watchdog traffic are all things only that sweep exercises.

`Closes #167` will not fire from a squash merge onto a non-default base, so #167 needs closing by hand.


cyon1c:
I'm inclined to suggest a small edit to this - can we make Awaken/sleep/idle/supporter ready into a separate app from duel? This way we can use this "hub" app as our standard base, and then duel starts at DuelCountdown. Thoughts?

tire-fire:
Yea, that makes sense.

src/pdn/game/game-session.cpp
```diff
@@ -0,0 +1,394 @@
...
+            static_cast<GameSession*>(ctx)->onChainGameEventAckPacket(macAddress, data, dataLen);
+        },
+        this);
+    wirelessManager->setEspNowPacketHandler(
+        PktType::kChainConfirm,
+        [](const uint8_t* macAddress, const uint8_t* data, const size_t dataLen, void* ctx) {
+            static_cast<GameSession*>(ctx)->onChainConfirmPacket(macAddress, data, dataLen);
+        },
+        this);
+    wirelessManager->setEspNowPacketHandler(
```
cyon1c: Could we alter this method to take an array of packet types so that there isn't so much repeated code?

cyon1c: Maybe not the right move, just a curiosity.

tire-fire: Yea, I think we'll handle this in #233

cyon1c (approved):

